PR #4, opened 2018-07-06
[WIP] genericPrint - PragmaDirective

mattiaerre:
# Description

- print PragmaDirective

## Before
```
pragma   solidity   ^0.4.23;  
contract Inbox {
  string public message;

  constructor(string initialMessage) public {
    message = initialMessage;
  }

  function setMessage(string newMessage) public {
    message = newMessage;
  }
}




```

## After
```
pragma solidity ^0.4.23;
contract Inbox {
  string public message;

  constructor(string initialMessage) public {
    message = initialMessage;
  }

  function setMessage(string newMessage) public {
    message = newMessage;
  }
}

```


src/printer.js
```diff
@@ -1,4 +1,31 @@
 // https://prettier.io/docs/en/plugins.html#printers
-const genericPrint = (path, options) => options.originalText;
+const PRAGMA_DIRECTIVE = 'PragmaDirective';
+const SOURCE_UNIT = 'SourceUnit';
+const UNKNOWN_NODE_TYPE = 'UNKNOWN NODE TYPE';
```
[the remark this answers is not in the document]
mattiaerre: not really just personal pref; Ill use strings as you suggested as there's no real need of constants

src/printer.js
```diff
@@ -1,4 +1,31 @@
...
+  const lines = options.originalText.split('\n');
+  switch (node.type) {
+    case SOURCE_UNIT:
+      node.children.forEach(child => {
+        switch (child.type) {
+          case PRAGMA_DIRECTIVE:
+            if (child.loc.start.line === child.loc.end.line) {
+              lines[child.loc.start.line - 1] = `pragma ${child.name} ${
+                child.value
+              };`;
```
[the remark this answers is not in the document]
mattiaerre: where can I learn more about this?

src/printer.js
```diff
@@ -1,4 +1,31 @@
...
+const PRAGMA_DIRECTIVE = 'PragmaDirective';
+const SOURCE_UNIT = 'SourceUnit';
+const UNKNOWN_NODE_TYPE = 'UNKNOWN NODE TYPE';
+
+const genericPrint = (path, options) => {
+  const node = path.getValue();
+  const lines = options.originalText.split('\n');
+  switch (node.type) {
+    case SOURCE_UNIT:
+      node.children.forEach(child => {
```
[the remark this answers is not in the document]
mattiaerre: where can I learn more about this?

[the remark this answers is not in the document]
j-f1: Prettier has some [docs](https://github.com/prettier/prettier/blob/master/commands.md) about the doc-printer commands.

src/printer.js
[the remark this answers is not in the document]
j-f1: See my other comment. In this case, you could do something like this:

```js
concat([
  "pragma",
  indent(concat([line, child.name, line, child.value])),
  ";"
]);
```

__tests__/node.json
```diff
@@ -0,0 +1,223 @@
+{
```
mattiaerre: this is the AST output

src/parser.js
```diff
@@ -1,9 +1,6 @@
+// https://prettier.io/docs/en/plugins.html#parsers
 const parser = require('solidity-parser-antlr');
 
-// https://prettier.io/docs/en/plugins.html#parsers
-// eslint-disable-next-line arrow-body-style
-const parse = text => {
-  return parser.parse(text);
-};
+const parse = text => parser.parse(text, { loc: true });
```
mattiaerre: this is in order to get location info for every node

README.md
```diff
@@ -2,12 +2,18 @@
...
-## Use
+## Resources
+
+- https://github.com/mattiaerre/prettier-plugin-solidity/issues/2
+
+## Test
 
 ```bash
-node_modules/prettier/bin-prettier.js --write **/*.sol --plugin=./node_modules/prettier-plugin-solidity
+node_modules/prettier/bin-prettier.js --write **/*.sol --plugin=./src
```
j-f1: This should be `node_modules/.bin/prettier` — the path to the actual CLI could change at any time.

README.md
```diff
@@ -2,12 +2,18 @@
...
 ```
 
-## Resources
+## Use
 
-- https://github.com/mattiaerre/prettier-plugin-solidity/issues/2
+```bash
+node_modules/prettier/bin-prettier.js --write **/*.sol --plugin=./node_modules/prettier-plugin-solidity
```
j-f1: Same here.

src/printer.js
```diff
@@ -1,4 +1,46 @@
...
+                child.name,
+                space,
+                child.value,
+                ';',
+                line,
+                line
+              ])
+            );
+            break;
+          default:
```
mattiaerre: I will add more and more cases and recursively call print to handle them all while default should not change too much I hope 😅

__tests__/node.json
j-f1: Is there a reason you’re checking this into the repository? I’d recommend doing tests with snapshots (see [`tests/arrays`](https://github.com/prettier/prettier/tree/master/tests/arrays) in the Prettier codebase for an example), and only checking that the output matches your expected output.

__tests__/lines.js
```diff
@@ -0,0 +1,15 @@
+module.exports = [
```
j-f1: This shouldn’t be necessary — your plugin shouldn’t need to split the text into lines to process it, and you should try to stick to just reading from the AST whenever possible.

src/printer.js
```diff
@@ -1,4 +1,46 @@
...
+            parts.push(
+              concat([
+                'pragma',
+                space,
+                child.name,
+                space,
+                child.value,
+                ';',
+                line,
+                line
```
j-f1: `line` will print a space instead of a newline if whatever comes after will fit on the same line. I think you want `hardline` here, since that will always introduce a line break.

src/printer.js
```diff
@@ -1,4 +1,46 @@
...
+              ])
+            );
+            break;
+          default:
+            for (
+              let index = child.loc.start.line;
+              index <= child.loc.end.line;
+              index += 1
+            ) {
+              parts.push(concat([lines[index - 1], line]));
```
j-f1: Same thing here with `line` vs. `hardline`.

__tests__/node.json
mattiaerre: yes definitely I'll add jest sooner but I wanted to have a ref of an AST just while I'm getting familiar w/ that structure as well as the prettier plugin architecture; this will be removed for sure

src/printer.js
```diff
@@ -1,4 +1,46 @@
...
+            }
```
j-f1: I feel like this would be better implemented in this way:

```js
parts.push(options.originalText.slice(child.range[0], child.range[1]))
```

(this requires `range: true` to be passed to the parser)

src/printer.js
```diff
@@ -1,4 +1,46 @@
...
+const space = ' ';
+
+const genericPrint = (path, options) => {
+  const node = path.getValue();
+  const lines = options.originalText.split('\n');
+  const parts = [];
+
+  switch (node.type) {
+    case 'SourceUnit':
+      node.children.forEach(child => {
```
j-f1: To avoid this nesting, you can move the `PragmaDirective` case into the parent `switch` statement, then call `path.map` in the `SourceUnit` printer as I suggested earlier.

README.md
mattiaerre: of course, thanks for that!

__tests__/lines.js
mattiaerre: I'll remove this as per my prev comment

src/printer.js
```diff
@@ -1,4 +1,30 @@
...
+        'pragma',
+        space,
+        node.name,
+        space,
+        node.value,
+        ';',
+        hardline
+      ]);
+    default:
+      return concat([
```
mattiaerre: at the moment this default returns the source code as it is, I've had to add a `+ 1` to the end index and a `hardline` at the very end

src/printer.js
```diff
@@ -1,4 +1,30 @@
...
+const { concat, hardline, join } = require('prettier').doc.builders;
+
+const space = ' ';
+
+const genericPrint = (path, options, print) => {
+  const node = path.getValue();
+
+  switch (node.type) {
+    case 'SourceUnit':
+      return join(hardline, path.map(print, 'children'));
```
mattiaerre: as it is `children` containing `PragmaDirective` and other node types; so if I understood properly, the `map` function invokes the print function against every element in the children collection?

src/printer.js
```diff
@@ -1,4 +1,26 @@
...
+    case 'PragmaDirective':
+      return concat([
+        join(space, ['pragma', node.name, node.value]),
+        ';',
+        hardline
+      ]);
```
mattiaerre: or maybe better:

```javascript
      return concat([
        'pragma',
        space,
        node.name,
        space,
        node.value,
        ';',
        hardline
      ]);
```

src/printer.js
```diff
@@ -1,4 +1,26 @@
 // https://prettier.io/docs/en/plugins.html#printers
-const genericPrint = (path, options) => options.originalText;
+const { concat, hardline, join } = require('prettier').doc.builders;
+
+const space = ' ';
```
mattiaerre: is there a better way to deal w/ spaces?

src/printer.js
j-f1: That’s correct! One pitfall to be aware of is that there’s only one `path` object that gets mutated when you call `path.map`/`path.call` and passed to `print`.

src/printer.js
j-f1: Awesome! This will start being used less and less as you finish the parser. 😎

src/printer.js
j-f1: Generally you’re going to want to use `line` and `indent()` if the language will allow you to insert a newline at that specific place. In the cases (rare for JS, perhaps not so much for other languages) where you can’t break, using a literal `' '` can help make your intentions clear.

src/printer.js
mattiaerre: nope it is not; pragma should always be in one line;

src/printer.js
```diff
@@ -1,4 +1,33 @@
...
+  switch (node.type) {
+    case 'SourceUnit':
+      return join(hardline, path.map(print, 'children'));
+    case 'PragmaDirective':
+      return concat([
+        join(space, ['pragma', node.name, node.value]),
+        ';',
+        hardline
+      ]);
+    case 'ContractDefinition':
```
mattiaerre: I've added contract definition but I've got a very hard time w/ `hardline` and indentation ... not sure what's wrong w/ this case here 😢

src/clean.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+function clean(ast, newObj, parent) {
+  ['range', 'loc'].forEach(name => {
+    delete newObj[name];
+  });
+}
+
+module.exports = clean;
```
mattiaerre: do you mind if I change this into:

```javascript
  ['loc', 'range'].forEach(name => {
    if (newObj[name]) {
      delete newObj[name]; // eslint-disable-line no-param-reassign
    }
  });
};

module.exports = clean;
```

I am not a big fan of param reassign and unused vars, is it possible to change the signature of this function?

j-f1: You can delete the `parent` parameter since we don’t use it now, but I’m curious why you are checking if the key exists before deleting it. Shouldn’t it be harmless to delete something that doesn’t exist?

mattiaerre: yes my bad ... 😊 you are right no need to check; we can keep the parameter there if you think we may be need it later

mattiaerre:
as I moved this repo the last commit is only in order to check the new Travis CI integration:

https://github.com/prettier-solidity/prettier-plugin-solidity/pull/4/commits/fc4ac627934807f9731e883fe3e95b5d69854c4b

mattiaerre:
shall we merge this or wait for @federicobond to integrate his `print` function? another option would be to open a new PR w/ that component only, what's your preference?

// cc @j-f1

j-f1:
Let’s merge this now so it doesn’t get any larger, then we can have @federicobond make PRs for any additions.

federicobond:
Let’s keep the momentum going. As long as code is simple to understand/change, we can continue iterating to the correct solution.

PR #5, opened 2018-07-07
Add tests

j-f1:
I’ve added Jest (with support files copied from the Prettier codebase) to easily test how files are printed and how they change over time, and added commands to run Prettier on the codebase.

If you want to enable Travis CI to ensure code is formatted properly (both JS and Solidity), you can start [here](https://github.com/marketplace/travis-ci/order/MDIyOk1hcmtldHBsYWNlTGlzdGluZ1BsYW43MA==).

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,0 +1,21 @@
+"use strict";
+
+const ENABLE_COVERAGE = !!process.env.CI;
+
+const semver = require("semver");
+const isOldNode = semver.parse(process.version).major <= 4;
+
+module.exports = {
+  setupFiles: ["<rootDir>/tests_config/run_spec.js"],
+  snapshotSerializers: ["<rootDir>/tests_config/raw-serializer.js"],
+  testRegex: "jsfmt\\.spec\\.js$|__tests__/.*\\.js$",
+  collectCoverage: ENABLE_COVERAGE,
+  collectCoverageFrom: ["src/**/*.js", "!<rootDir>/node_modules/"],
+  testEnvironment: "node",
+  transform: {},
+  watchPlugins: [
+    "jest-watch-typeahead/filename",
+    "jest-watch-typeahead/testname"
+  ]
+};
+
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,15 +1,10 @@
 'use strict';
 
-const ENABLE_COVERAGE = !!process.env.CI;
-
-const semver = require('semver');
-const isOldNode = semver.parse(process.version).major <= 4;
-
 module.exports = {
   setupFiles: ['<rootDir>/tests_config/run_spec.js'],
   snapshotSerializers: ['<rootDir>/tests_config/raw-serializer.js'],
   testRegex: 'jsfmt\\.spec\\.js$|__tests__/.*\\.js$',
-  collectCoverage: ENABLE_COVERAGE,
+  collectCoverage: !!process.env.CI,
   collectCoverageFrom: ['src/**/*.js', '!<rootDir>/node_modules/'],
   testEnvironment: 'node',
   transform: {},
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,21 +1,20 @@
-"use strict";
+'use strict';
 
 const ENABLE_COVERAGE = !!process.env.CI;
 
-const semver = require("semver");
+const semver = require('semver');
 const isOldNode = semver.parse(process.version).major <= 4;
 
 module.exports = {
-  setupFiles: ["<rootDir>/tests_config/run_spec.js"],
-  snapshotSerializers: ["<rootDir>/tests_config/raw-serializer.js"],
-  testRegex: "jsfmt\\.spec\\.js$|__tests__/.*\\.js$",
+  setupFiles: ['<rootDir>/tests_config/run_spec.js'],
+  snapshotSerializers: ['<rootDir>/tests_config/raw-serializer.js'],
+  testRegex: 'jsfmt\\.spec\\.js$|__tests__/.*\\.js$',
   collectCoverage: ENABLE_COVERAGE,
-  collectCoverageFrom: ["src/**/*.js", "!<rootDir>/node_modules/"],
-  testEnvironment: "node",
+  collectCoverageFrom: ['src/**/*.js', '!<rootDir>/node_modules/'],
+  testEnvironment: 'node',
   transform: {},
   watchPlugins: [
-    "jest-watch-typeahead/filename",
-    "jest-watch-typeahead/testname"
+    'jest-watch-typeahead/filename',
+    'jest-watch-typeahead/testname'
   ]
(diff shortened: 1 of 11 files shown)


j-f1:
If you have any questions about how this all works, feel free to ask!

mattiaerre:
this is awesome! I'll go through this PR later and add comments/ask questions about it. thanks; I really appreciate your help @j-f1

package.json
```diff
@@ -4,7 +4,10 @@
   "description": "prettier plugin for solidity",
   "main": "src",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "prettier": "prettier \"{.,{src,tests_config}/**}/*.{js{,on},md}\" \"tests/**/jsfmt.spec.js\"",
+    "lint": "npm run prettier -- --list-different",
```
mattiaerre: interesting, are we using prettier as a linter as well?

package.json
```diff
@@ -30,7 +33,9 @@
     "eslint-config-airbnb": "^17.0.0",
     "eslint-plugin-import": "^2.13.0",
     "eslint-plugin-jsx-a11y": "^6.1.0",
-    "eslint-plugin-react": "^7.10.0"
+    "eslint-plugin-react": "^7.10.0",
+    "jest": "^23.3.0",
+    "jest-watch-typeahead": "^0.1.0"
```
mattiaerre: this is really cool 👍

tests/pragma/jsfmt.spec.js
```diff
@@ -0,0 +1 @@
+run_spec(__dirname);
```
mattiaerre: can you explain this a little bit more please?

tests/pragma/solidity.sol
```diff
@@ -0,0 +1,3 @@
+pragma solidity ^0.4.23;
+pragma   solidity   ^0.4.23;
+
```
mattiaerre: this should be an example of solidity code?

tests_config/raw-serializer.js
```diff
@@ -0,0 +1,17 @@
+// source: https://github.com/prettier/prettier/blob/ee2839bacbf6a52d004fa2f0373b732f6f191ccc/tests_config/raw-serializer.js
+'use strict';
+
+const RAW = Symbol.for('raw');
+
+module.exports = {
```
mattiaerre: this is cool! why do we want to use a custom serializer? so that we don't have any additional "noise" from the standard one?

tests_config/run_spec.js
```diff
@@ -0,0 +1,125 @@
+// source: https://github.com/prettier/prettier/blob/ee2839bacbf6a52d004fa2f0373b732f6f191ccc/tests_config/run_spec.js
```
mattiaerre: can you explain me a little bit more about this, please?

package.json
j-f1: Yep! It can’t hurt :)

jest.config.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+const ENABLE_COVERAGE = !!process.env.CI;
+
+const semver = require('semver');
+const isOldNode = semver.parse(process.version).major <= 4;
```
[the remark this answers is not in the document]
j-f1: Fixed!

j-f1:
Here’s how the `tests` directory works:

1. Create a directory inside it. You can name it whatever you want, but it’s usually useful to name it based on the constructs you’ll be testing.
2. Create a `jsfmt.spec.js` file inside that new directory, and put `run_spec(__dirname);` in it. This will tell Jest to look in directory, and the `run_spec.js` file will run the necessary tests.
3. Create as many `.sol` files in the new directory as you want. Each one will be run through your plugin, and the output will be stored in a `__snapshots__/jsfmt.spec.js.snap` file that will be automatically created when you run `npm test`.
4. If you change the snapshot files or the printer, you’ll need to run `npm test -- -u` to update the snapshots to show the new output.

The `raw-serializer` serializer will output the snapshotted string directly into the snapshot file instead of wrapping it with quotes, which would then necessitate escaping any quotes in the outputted snapshot, making it somewhat harder to read.

jest.config.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+  setupFiles: ['<rootDir>/tests_config/run_spec.js'],
+  snapshotSerializers: ['<rootDir>/tests_config/raw-serializer.js'],
+  testRegex: 'jsfmt\\.spec\\.js$|__tests__/.*\\.js$',
```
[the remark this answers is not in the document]
j-f1: IMO the `__tests__` directory is useful for testing any utilities you may create — the `jsfmt.spec.js` stuff I explain in my comment below ⬇️ is much more efficient for testing the printer as a whole.

tests/pragma/jsfmt.spec.js
j-f1: See my comment below ⬇️

tests/pragma/solidity.sol
j-f1: Yep! See my comment below for more details ⬇️

tests_config/raw-serializer.js
j-f1: That’s correct! Otherwise, you’d get something like this in the `.snap` file:

```js
exports[`solidity.sol 1`] = `
"pragma solidity ^0.4.23;
pragma   solidity   ^0.4.23;

pragma solidity ^0.4.23;

pragma solidity ^0.4.23;

"
`;
```

j-f1:
The snapshots in the snapshot files look like this:

```js
exports[`solidity.sol 1`] = `
pragma solidity ^0.4.23;
pragma   solidity   ^0.4.23;

pragma solidity ^0.4.23;

pragma solidity ^0.4.23;

`;
```

The first part (above the line of `~`) is the input code. We keep a copy of this so that if the test file changes but the output doesn’t, we’ll be alerted that something might be wrong. The line of `~` indicates the specified `printWidth` (you can override this per-directory by using `run_spec(__dirname, { printWidth: whatever });` in your `jsfmt.spec.js`, and you can run `run_spec` as many times as you’d like with whatever options you want). The rest of the snapshot (below the `~` is your plugin’s output.

j-f1:
Once this is merged, I’ll see what I can do about `ContractDefinition` and your troubles with `hardline`.

mattiaerre:
thanks for explaining the infrastructure; I've added Travis-CI as you can see and these are the env variables I set based on your PR:

```dotenv
AST_COMPARE=true
CI=true
TEST_STANDALONE=true
```

as soon as everything is 💚 I'll merge it!

j-f1:
I don’t think you need the `CI=true` variable — Travis should be setting it automatically.

mattiaerre:
I'll remove CI=true and see what happens 😅

PR #9, opened 2018-07-10
housekeeping

mattiaerre:
# Description

- set `run_spec` as global (eslint)

- fix package.json [WIP]

- remove 'use strict'

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,8 @@
 {
   "extends": "airbnb",
+  "globals": {
+    "run_spec": false
+  },
   "rules": {
     "arrow-parens": "off",
     "comma-dangle": "off",
--- a/package.json
+++ b/package.json
@@ -4,10 +4,11 @@
   "description": "prettier plugin for solidity",
   "main": "src",
   "scripts": {
-    "prettier": "prettier \"{.,{src,tests_config}/**}/*.{js{,on},md}\" \"tests/**/jsfmt.spec.js\"",
+    "coverage": "codecov",
     "eslint": "eslint src/**",
     "lint": "npm run eslint && npm run prettier -- --list-different",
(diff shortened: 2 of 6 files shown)


mattiaerre:
In addition, it'd be nice if:

1. @federicobond and @j-f1 you could fix your info as contributors in the package.json file

2. see if `raw-serializer.js` and `run_spec.js` could be consumed as modules; not sure if they are already available but it'd be cool to extract them from prettier; what do you think?

.eslintrc
```diff
@@ -1,5 +1,8 @@
 {
```
j-f1: Can you add ESLint to the project explicitly (i.e. so `npm run lint` will run it?

mattiaerre: oh snap! 😊 I'll do it!

package.json
```diff
@@ -5,13 +5,14 @@
   "main": "src",
   "scripts": {
     "prettier": "prettier \"{.,{src,tests_config}/**}/*.{js{,on},md}\" \"tests/**/jsfmt.spec.js\"",
+    "eslint": "eslint src/**",
```
mattiaerre: I would keep all the scripts @j-f1 what do you think?

j-f1: Yeah let’s keep these, but can you add `npm run eslint` to the `lint` script and `npm run eslint -- --fix` to the `lint:fix` script?

mattiaerre:
hi @federicobond I've added your email and if the CI is 💚 I'm going to merge this PR; let me know if you want me to use a different email; also re your print function I'll be more than happy to integrate it myself using your gist if you like and you are busy; let me know what you prefer. thanks so much,

j-f1:
> @j-f1 shall we merge?

SGTM 👍

PR #21, opened 2018-07-31
printer

mattiaerre:
# Description

- add @federicobond 's `genericPrint` function


tests/contract/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -17,11 +17,12 @@ contract Inbox {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 pragma solidity ^0.4.23;
 contract Inbox {
-  string public message;
+  string message;
```
mattiaerre: I think we should preserve the `public` keyword right?

mattiaerre:
there is an issue w/ TravisCI; I've sent an email to support

mattiaerre:
I've been able to reproduce it locally running the following command:

`AST_COMPARE=true npm t`

I'd love to have your POV on this @j-f1 and @federicobond 
thanks,

src/printer.js
```diff
@@ -254,7 +254,7 @@ function genericPrint(path, options, print) {
       if (node.isIndexed) {
         doc = join(' ', [doc, 'indexed']);
       }
-      return join(' ', [doc, node.name]);
+      return join(' ', [doc, node.visibility, node.name]);
```
mattiaerre: what do you think about this @federicobond ?

mattiaerre: this line above fixes also the tests; i think this is good!

PR #39, opened 2018-09-21
space-after-if-condition

### LINKED ISSUE #37: Add space before `if`'s opening braces
Output right now:

```
if (condition){
```

Desired output:

```
if (condition) {
```

### LINKED ISSUE #38: Put `while` opening brace in the same line
Current output:

```
while (condition)
{
  ...
```

Desired output:

```
while (condition) {
  ...
```

mattiaerre:
# Description

- add a line before `ContractDefinition` 

- add space after the condition for `IfStatement`

- fix #37 

- fix #38

### DIFF
--- a/tests/BasicIterator/BasicIterator.sol
+++ b/tests/BasicIterator/BasicIterator.sol
@@ -1,0 +1,44 @@
+/*
+	This is a very simple demonstration of a while loops. Same as JS/c.
+*/
+
+contract BasicIterator {
+
+    address creator;                  // reserve one "address"-type spot
+    uint8[10] integers;               // reserve a chunk of storage for 10 8-bit unsigned integers in an array
+
+    function BasicIterator() 
+    {
+        creator = msg.sender;         // set the creator address
+        uint8 x = 0;                  // initialize an 8-bit, unsigned integer to zero
+        while(x < integers.length)    // the variable integers was initialized to length 10
+        {
+        	integers[x] = x;      // set integers to [0,1,2,3,4,5,6,7,8,9] over ten iterations
+        	x++;
+        }
+    }
+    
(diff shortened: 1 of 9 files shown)


mattiaerre:
it fails if you run `AST_COMPARE=true npm run test`; can you help us w/ this @j-f1 ?

j-f1:
Those are the kinds of issues that the AST-massaging function can handle.

mattiaerre:
not sure if I understand what you mean; shall we change something here?

https://github.com/prettier-solidity/prettier-plugin-solidity/blob/master/tests_config/run_spec.js#L52-L71

j-f1:
I *think* so.

tests/BasicIterator/BasicIterator.sol
```diff
@@ -7,7 +7,7 @@ contract BasicIterator {
     address creator;                  // reserve one "address"-type spot
     uint8[10] integers;               // reserve a chunk of storage for 10 8-bit unsigned integers in an array
 
-    function BasicIterator()
```
mattiaerre: dah 😅 @j-f1 as we "translate" `function ContractName` => `constructor`

fvictorio: The plugin is doing that right now? I mean, converting `function Contract` to `constructor`? Because that shouldn't be prettier's responsibility IMO.

mattiaerre: 100% agree. Apparently it does that. See if I can fix it.

fvictorio: I guess the parser adds the `isConstructor` metadata to functions named as the contract, and the printer function uses that to print `constructor`.

PR #40, opened 2018-09-26
preserve-comments

mattiaerre:
# Description

- something working !?!


fvictorio:
I added some changes:

- Use the `raw` property in block comments (otherwise, the extra `*` were being lost)
- Remove all unnecessary code
- Remove some properties from the parsed comments that were causing `AST_COMPARE=true npm test` to fail.

src/parser.js
```diff
@@ -1,6 +1,11 @@
+const extract = require('extract-comments');
 // https://prettier.io/docs/en/plugins.html#parsers
 const parser = require('solidity-parser-antlr');
 
-const parse = text => parser.parse(text, { loc: true, range: true });
+const parse = text => {
+  const parsed = parser.parse(text, { loc: true, range: true });
+  parsed.comments = extract(text);
```
mattiaerre: this is awesome, thanks for that @fvictorio

src/clean.js
```diff
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-unused-vars
 const clean = (ast, newObj, parent) => {
-  ['loc', 'range'].forEach(name => {
+  ['code', 'codeStart', 'loc', 'range'].forEach(name => {
```
mattiaerre: this function is used only when `AST_COMPARE=true`

src/parser.js
```diff
@@ -5,6 +5,15 @@ const parser = require('solidity-parser-antlr');
...
   const parsed = parser.parse(text, { loc: true, range: true });
   parsed.comments = extract(text);
+
+  // `code` and `codeStart` are different after prettifying, so the tests fail when they have AST_COMPARE enabled.
+  // Since we don't use those properties, we just remove them.
+  parsed.comments.forEach(comment => {
+    delete comment.code;
+    delete comment.codeStart;
+    return comment;
+  });
```
j-f1: This kind of code should go in a `massageAST` function so you don’t have to iterate over all the comments again when formatting with `AST_COMPARE` enabled.

src/clean.js
mattiaerre: yes @j-f1 this is what we've done thanks 😉

PR #43, opened 2018-09-27
moar-on-empty-lines

### LINKED ISSUE #41: Don't add empty lines between documenting comments and functions
#40 makes the plugin preserve comments, but creates a new problem. If you have some documentation above a function:

```solidity
/*
 * This function does stuff.
 */
function doStuff() public {
  ...
}
```

The prettified version will be:

```solidity
/*
 * This function does stuff.
 */

function doStuff() public {
  ...
}
```

The cause of this is that we prepend function definitions with empty lines. Possible solutions:

1. Improve the `printer` function to separate functions with empty lines in a more intelligent way (for example, append an empty line to all function definitions except the last).
2. Remove that feature completely. Prettier does not separate functions in javascript classes —maybe we shouldn't either, and this should be a responsibility of a linter.

mattiaerre:
# Description

- add lines at the end

- fix #41

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -15,7 +15,7 @@ function genericPrint(path, options, print) {
     case 'SourceUnit':
       return join(hardline, path.map(print, 'children'));
     case 'PragmaDirective':
-      return concat(['pragma ', node.name, ' ', node.value, ';']);
+      return concat(['pragma ', node.name, ' ', node.value, ';', line]);
     case 'ImportDirective':
       // @TODO: handle proper escaping
       doc = concat(['"', node.path, '"']);
@@ -44,7 +44,6 @@ function genericPrint(path, options, print) {
         ]);
       }
       return concat([
-        line,
         doc,
         ' {',
         indent(line),
@@ -97,7 +96,7 @@ function genericPrint(path, options, print) {
         ]);
       }
       if (node.body) {
-        return concat([line, join(' ', [doc, path.call(print, 'body')])]);
+        return concat([join(' ', [doc, path.call(print, 'body')]), line]);
       }
       return concat([doc, ';']);
     case 'ParameterList':
--- a/tests/BasicIterator/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/BasicIterator/__snapshots__/jsfmt.spec.js.snap
@@ -50,11 +50,9 @@ contract BasicIterator {
 	This is a very simple demonstration of a while loops. Same as JS/c.
 */
 
-
 contract BasicIterator {
   address creator; // reserve one "address"-type spot
   uint8[10] integers; // reserve a chunk of storage for 10 8-bit unsigned integers in an array
-
   function BasicIterator() {
     creator = msg.sender; // set the creator address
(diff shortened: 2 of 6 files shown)


mattiaerre:
Hello, we really need some help here re how to treat empty lines, new lines, and spaces in general. Any input about this subject would be really really appreciated; we have in place some solution that adds `line` after every function and before the contract definition but clearly, we are not doing it right and it feels we'll never find a decent compromise; what would you suggest us to do? Thanks so very much!

// cc @j-f1 @vjeux @ikatyang @azz @duailibe

vjeux:
The way we're doing it with prettier is to change the way we display "sequences" and whenever we're going to insert a newline, check if the next line is empty and if it is, then insert a hard line. https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L1188-L1190

We only preserve lines (at most one) in the middle of the sequence, not before the first element or after the last one.

This has been a good balance between being restrictive and yet giving freedom to developers.

mattiaerre:
thank you so much @vjeux; is prettier exporting that `isNextLineEmpty` helper function? this is an awesome tip and I believe @fvictorio was already looking into that; we are really eager to launch this plugin in a decent state but we need to finalize a little bit better this issue; so thanks for the quick response!

vjeux:
I don't know... If we don't currently, it probably is a good idea to. I don't have the time to work on it but if you send a PR on prettier itself, it'll likely be merged in. Otherwise the implementation is only a few lines.

mattiaerre:
it seems available @vjeux 

```javascript
const { isNextLineEmpty } = require('prettier').util;
```

// cc @fvictorio

fvictorio:
@vjeux Thanks a lot for the help!

I played with this a little and I think that approach could work. The problem that I found is this: in prettier's parser (at least the one that uses by default), the `locEnd` is the character _after_ the closing brace:

```javascript
function foo () {
}|
```

With our parser, the `locEnd` is the closing brace.

If I understand it correctly, the way `isNextLineEmpty` works is by skipping characters from the `locEnd` until the next line, and then checking if it's empty. If it starts with the closing brace, it doesn't ignore any character and so it returns false. (One of the characters that it ignores is the semicolon, so this approach does work for state variables and events).

Some possible approaches:

1. Use `isNextLineEmptyAfterIndex` instead, and pass it `locEnd + 1` as the starting index
2. Write our own `isNextLineEmpty`
3. Modify the parser

1 is the easiest way, although I'm not sure if that would give us problems in edge cases. I would do it anyway, at least to have something working with passing tests; we can refactor later.

vjeux:
Ohhh, that's annoying. A lot of the comment/newline logic is based on the fact that locEnd is after the closing element. I would recommend adding a small pass after the parser to normalize this. Otherwise it's going to cause a lot of subtle problems down the line.

You may also want to ask whoever wrote the parser to fix it in their code. This is likely going to be an issue for the entire ecosystem of tools built on-top of it.

mattiaerre:
I'm also down to close this in favor of #44 

// cc @fvictorio

mattiaerre:
as per my last comment

PR #48, opened 2018-10-02
Preserve constant keyword in variable declarations

fvictorio:


src/printer.js
```diff
@@ -260,12 +260,13 @@ function genericPrint(path, options, print) {
       if (node.isIndexed) {
         doc = join(' ', [doc, 'indexed']);
       }
+      const constantKeyword = node.isDeclaredConst ? 'constant' : '';
```
mattiaerre: instead of adding the rule to `.eslint` shall we exclude it explicitly here? something like:

```javascript
      // eslint-disable-next-line no-case-declarations
      const constantKeyword = node.isDeclaredConst ? 'constant' : '';
      if (node.visibility === 'default') {
        return join(
          ' ',
          [doc, constantKeyword, node.name].filter(element => element)
        );
      }
      return join(
        ' ',
        [doc, node.visibility, constantKeyword, node.name].filter(
          element => element
        )
      );
```

mattiaerre: also, the PR fails because you need to run prettier against the code 😊 
`npm run lint`

fvictorio: Oh, I just made a comment about this [here](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/47#discussion_r223181124)

fvictorio: > also, the PR fails because you need to run prettier against the code

Fixed, thanks.

mattiaerre: shall we wrap the `'VariableDeclaration'` in a block as well and get rid of the rule in the `.eslintrc` file?

fvictorio:
> happy to merge this @fvictorio if you can address my last comment please

Done!

PR #68, opened 2018-10-21
Print else on new line when consequent is not a block

### LINKED ISSUE #50: If statement in one line is not being prettified
A contract like this:

```solidity
contract test {
  function fun(uint256 a) returns (address b) {
    if (a < 0) b = 0x67; else if (a == 0) b = 0x12; else b = 0x78;
  }
}
```

is prettified like this:

```solidity
contract test {
  function fun(uint256 a) returns(address b) {
    if (a < 0) b = 0x67; else if (a == 0) b = 0x12; else b = 0x78;
  }
}
```

Almost no changes.

I'm not sure what is the best output here. Should we add the braces or is that too opinionated?

fvictorio:
Closes #50.

I took "inspiration" (AKA stole it) from here:

https://github.com/prettier/prettier/blob/559831d08eadbda644ba6299a562c245316fa2e4/src/language-js/printer-estree.js#L1624-L1625

Notice that they also check something related to the comments (not sure what), but that's something we can do in another issue if someone reports it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "prettier-plugin-solidity",
-  "version": "1.0.0-alpha.7",
+  "version": "1.0.0-alpha.9",
   "description": "prettier plugin for solidity",
   "main": "src",
   "scripts": {
--- a/src/printer.js
+++ b/src/printer.js
@@ -325,7 +325,13 @@ function genericPrint(path, options, print) {
         path.call(print, 'trueBody')
       ]);
       if (node.falseBody) {
-        doc = concat([doc, ' else ', path.call(print, 'falseBody')]);
+        const elseOnSameLine = node.trueBody.type === 'Block';
+        doc = concat([
+          doc,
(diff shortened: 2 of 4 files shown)


j-f1:
The comment check is to print a comment before `else` properly.

mattiaerre:
very nice thanks @fvictorio what happens if we add a comment before the `else`?

fvictorio:
> what happens if we add a comment before the `else`?

```solidity
contract test {
  function fun(uint256 a) returns (uint) {
    if (something) foo();
    // comment
    else if (somethingElse) bar();
    else whatever();
  }
}
```

gets formatted as

```solidity
contract test {
  function fun(uint256 a) returns(uint) {
    if (something) foo();
    else // comment
    if (somethingElse) bar();
    else whatever();
  }
}
```

FWIW, this also happens if the first `if` body has braces.

I think we should merge this and deal with that in another PR.

PR #69, opened 2018-10-22
Omit semicolon in for statements init and loop expressions

### LINKED ISSUE #64: issue w/ for statement
# Description

missing a semicolon if for statement body is not included in braces.

## Source

```javascript
  function count(string[] data) public pure returns (uint) {
    uint result;
    for(uint i = 0; i < data.length; i++)
      result += 1;
    return result;
  }
```

## Prettified

```javascript
  function count(string[] data) public pure returns(uint) {
    uint result;
    for (uint i = 0; i < data.length; i++) result += 1
    return result;
  }
```

> N.B. there is a missing `;` at the end of the for body

fvictorio:
Closes #64.

I ended up mutating the nodes anyway, but I did it in the parser, which make it a little less ugly (or so I tell myself). The problem is that we don't have a better `type` to assign to this nodes; at least I didn't see one that could work.

Maybe later we can do it in a cleaner way. For now, tests pass and I don't think it's a huge technical debt.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -162,10 +162,9 @@ function genericPrint(path, options, print) {
       ]);
 
     case 'ExpressionStatement': {
-      const addSemicolon = path.getParentNode().type !== 'ForStatement';
       return concat([
         node.expression ? path.call(print, 'expression') : '',
-        addSemicolon ? ';' : ''
+        node.omitSemicolon ? '' : ';'
       ]);
     }
     case 'FunctionCall':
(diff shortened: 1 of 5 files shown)


src/printer.js
```diff
@@ -245,8 +244,7 @@ function genericPrint(path, options, print) {
       if (node.initialValue) {
         doc = concat([doc, ' = ', path.call(print, 'initialValue')]);
       }
-      const addSemicolon = path.getParentNode().type !== 'ForStatement';
-      return concat([doc, addSemicolon ? ';' : '']);
+      return concat([doc, node.omitSemicolon ? '' : ';']);
```
mattiaerre: and it works also here! this is fantastic 👍

Issue #70, opened 2018-10-22
Comments before else are not formatted correctly


fvictorio:
```solidity
contract test {
  function fun(uint256 a) returns (uint) {
    if (something) foo();
    // comment
    else if (somethingElse) bar();
    else whatever();
  }
}
```

gets formatted as

```solidity
contract test {
  function fun(uint256 a) returns(uint) {
    if (something) foo();
    else // comment
    if (somethingElse) bar();
    else whatever();
  }
}
```

This also happens if the first `if` body has braces.

_Originally posted by @fvictorio in https://github.com/prettier-solidity/prettier-plugin-solidity/pull/68/comment#issuecomment-432027484_

PR #71, opened 2018-10-31
prettier-comments

mattiaerre:
# Description

this PR is to show how to integrate and use `prettier-comments`, have a look at [https://github.com/prettier-solidity/prettier-comments](https://github.com/prettier-solidity/prettier-comments) for more details

- relates to #70

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,4 @@
-const { handleComments } = require('prettier-comments'); // eslint-disable-line import/no-extraneous-dependencies
+const { handleComments } = require('./prettier-comments');
 
 const massageAstNode = require('./clean');
 const loc = require('./loc');
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,5 @@
+const { handleComments } = require('prettier-comments'); // eslint-disable-line import/no-extraneous-dependencies
+
 const massageAstNode = require('./clean');
 const loc = require('./loc');
 const parse = require('./parser');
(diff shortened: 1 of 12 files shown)


j-f1:
cc @prettier/core do you want to extract these to a package under @prettier?

tests/Etc/Etc.sol
```diff
@@ -32,4 +32,11 @@ contract Contract {
     uint sum;
     for ( i = 0; i < 10; i++ ) sum   +=  i;
   }
+
+  function fun(uint256 a) returns (uint) {
+    if (something) foo();
+    // comment
```
fvictorio: Shouldn't there be an updated snapshot for this too?

src/index.js
```diff
@@ -1,3 +1,5 @@
+const { handleComments } = require('prettier-comments'); // eslint-disable-line import/no-extraneous-dependencies
```
fvictorio: Any reason for the extraneous dependency?

fvictorio: Oh, nevermind, I see that you are using it linked for now.

tests/Etc/Etc.sol
mattiaerre: nope as it throws an error now as per:

https://github.com/prettier-solidity/prettier-comments#how-to-use-it-w-prettier-plugin-solidity

basically the `comments.js` I've extracted from `prettier` knows how to scan their AST but not the one we are using (this is my assumption)

mattiaerre:
No joy from the prettier team so I've decided to move the prettier-comments source into this project as I didn't want (not sure if I am allowed) to publish a separate module; I've managed to keep everything separate so that if there will be a module, we would just need to change the import; let me know what you think so that I can publish this version that fixes the comments issue.

// cc @j-f1 @fvictorio

tests/Etc/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -35,6 +35,13 @@ contract Contract {
     uint sum;
     for ( i = 0; i < 10; i++ ) sum   +=  i;
   }
+
+  function fun(uint256 a) returns (uint) {
+    if (something) foo();
+    // comment
```
mattiaerre: YAY!

PR #77, opened 2018-12-21
Fix prettier removing payable from address payable

svechinsky:
So I noticed an issue with the new address payable type.
At first the plugin would fail because the published version of solidity-parser-antlr had a bug and didn't recognize the payable keyword.
After using the most recent version from git I noticed that the printer would just remove the payable keyword.

So this PR fixes this.
I have absolutley no idea if what I did here is the correct way to fix it but it seems to work for me.
Also I have no idea what's the best way to include the most recent version of solidity-parser-antlr, maybe best to ask the maintainer to publish the newest version?

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -291,7 +291,12 @@ function genericPrint(path, options, print) {
       if (node.visibility === 'default') {
         return join(
           ' ',
-          [doc, node.typeName.stateMutability, constantKeyword, node.name].filter(element => element)
+          [
+            doc,
+            node.typeName.stateMutability,
+            constantKeyword,
+            node.name
+          ].filter(element => element)
         );
       }
       return join(
--- a/src/printer.js
+++ b/src/printer.js
@@ -135,7 +135,12 @@ function genericPrint(path, options, print) {
       doc = path.call(print, 'typeName');
       doc = join(
         ' ',
-        [doc, node.storageLocation, node.name].filter(element => element)
+        [
+          doc,
+          node.storageLocation,
+          node.typeName.stateMutability,
+          node.name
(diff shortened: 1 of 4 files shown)


svechinsky:
Fixed some styling issues, also https://github.com/federicobond/solidity-parser-antlr/issues/40 the new version was released so this should be mergable

fvictorio:
@svechinsky could you add tests for this?

mattiaerre:
thanks for this PR @svechinsky if you can add a test that would be really appreciated if you cannot please comment w/ an example from a contract where you've noticed the issue and we'll add the test for you as we'd like to make sure that the fix doesn't introduce regression as well as improve the overall code coverage

svechinsky:
I won't be able to write the test myself but will post an example later
today.

mattiaerre:
thanks @svechinsky

svechinsky:
So if you run prettier on this function the payable part will be removed (both in the parameters as well as in the declaration).
```solidity
contract AddressPayable{
 function sendSomeEth(address payable to){
   address payable target = to;
   target.transfer(100 eth);
}
}
```

mattiaerre:
hi @svechinsky can you grant me write access to your fork so that I can push the test? otherwise you can add the attached folder to the project under `tests`

[AddressPayable.zip](https://github.com/prettier-solidity/prettier-plugin-solidity/files/2715776/AddressPayable.zip)

fvictorio:
Added the tests. @mattiaerre, just FYI, you can push to a forked repo when there's a pull request like this :slightly_smiling_face:

fvictorio:
This closes #78, right?

PR #86, opened 2019-01-04
Add support for ThrowStatement

fvictorio:
This also adds a new test file called `AllSolidityFeatures.sol` that I ~stole~ borrowed from the 0x guys, specifically from [here](https://github.com/0xProject/0x-monorepo/blob/development/packages/sol-cov/test/fixtures/contracts/AllSolidityFeatures.sol).

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -509,6 +509,8 @@ function genericPrint(path, options, print) {
         ].filter(element => element)
       );
     }
+    case 'ThrowStatement':
+      return 'throw;';
     default:
       throw new Error(`Unknown type: ${JSON.stringify(node.type)}`);
   }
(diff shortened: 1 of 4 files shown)


fvictorio:
I don't know why Travis is failing :slightly_frowning_face: Tests are working locally for me.

mattiaerre:
@fvictorio you may want to run the test w/ this command:

`AST_COMPARE=true yarn test`

fvictorio:
@mattiaerre Thanks, I forgot about that. That exposed a lot of other problems, that I fixed in this PR:

- Things like `var (x) = 2;` and `var (x, y) = (2, 3);` were not being handled correctly.
- Modifiers without parentheses like `modifier myModifier {}` were printed as `modifier myModifier() {}` and the added parentheses apparently has a different AST.
- Base contracts with arguments like `contract Foo is Bar(0)` were being printed as `contract Foo is Bar`.

Issue #91, opened 2019-01-05
Add support for AssemblyFor


fvictorio:
May be related to #81.

`AssemblyFor` AST nodes are not supported.

PR #92, opened 2019-01-05
Add assembly case blocks

fvictorio:
Depends on #91.


mattiaerre:
@fvictorio can you resolve the conflicts so that I can merge? thanks

fvictorio:
@mattiaerre Fixed!

tests/Assembly/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -20,6 +32,18 @@ contract Assembly {
       }
     }
   }
+
+  function caseAssembly() {
+    assembly {
+      switch value
```
mattiaerre: uhm, shall we have 2 spaces to indent the "case" inside the switch?

fvictorio: You mean that the cases should be more indented than the switch? Yeah, I agree with that, and that's what prettier does in javascript.

fvictorio: Done.

mattiaerre: awesome! thanks so much

PR #94, opened 2019-01-06
Add support for AssemblyFor

### LINKED ISSUE #91: Add support for AssemblyFor

fvictorio:
Closes #91. Depends on #92.

I don't like how this is prettifying the fors, but at least now it won't fail when it finds them. This:

```
for { let i := 0 } lt(i, x) { i := add(i, 1) } { y := mul(2, y) }
```

is being printed as:

```
for {
  let i := 0
} lt(i, x) {
  i := add(i, 1)
} {
  y := mul(2, y)
}
```

I would prefer this:

```
for { let i := 0 } lt(i, x) { i := add(i, 1) } {
  y := mul(2, y)
}
```

(give or take).

The problem here is that the pre and post parts of the for are `AssemblyBlock`s, and in most cases it makes sense to print those with a newline after the opening brace.


tests/Assembly/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -20,6 +38,30 @@ contract Assembly {
...
+      }
+      case 1 {
+        mstore(0, 0x1111111111111111111111111111111111111111111111111111111111111111)
+      }
+    }
+  }
+
+  function forAssembly() {
+    assembly {
+      for {
```
mattiaerre: I do agree with you, this seems a little bit odd but that's a good starting point; I'll wait for you to address #92 and then we can merge both

mattiaerre:
some conflicts that need to be resolved @fvictorio

fvictorio:
@mattiaerre Done!

PR #95, opened 2019-01-06
assembly and ast

### LINKED ISSUE #81: issue w/ assembly code
as per @maxsam4 comment on Gitter:

Another bug: It did not refactor the assembly code in https://github.com/PolymathNetwork/polymath-core/blob/dev-3.0.0/contracts/proxy/Proxy.sol
it skipped a couple of parts

mattiaerre:
# Describe

- add test and clean

should fix #81 and #91 

maybe related to #94


src/clean.js
```diff
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-unused-vars
 const clean = (ast, newObj, parent) => {
-  ['code', 'codeStart', 'loc', 'range'].forEach(name => {
+  ['code', 'codeStart', 'loc', 'operations', 'range'].forEach(name => {
```
mattiaerre: this is in order to bypass the problem w/ `AST_COMPARE=true`

fvictorio: This is no longer necessary. Merging master into this branch fixes the issue.

PR #98, opened 2019-01-07
Improve function calls in assembly code

### LINKED ISSUE #97: Comments in assembly function calls are being merged
This

```solidity
contract Foo {
  function f() {
    g(
      1, // one
      2, // two
      3 // three
    );

    assembly {
      g(
        1, // one
        2, // two
        3 // three
      )
    }
  }
}
```

is being printed as:

```solidity
contract Foo {
  function f() {
    g(
      1, // one
      2, // two
      3 // three
    );

    assembly {
      g(1, 2, 3) // one // two // three
    }
  }
}
```

Notice that it seems to work fine for normal function calls; the problem seems to be inside assembly blocks.

fvictorio:
Closes #97.

### DIFF
--- a/src/printer.js
+++ b/src/printer.js
@@ -499,11 +499,20 @@ function genericPrint(path, options, print) {
       if (node.arguments.length === 0) {
         return node.functionName;
       }
-      // @TODO: add call args
       return concat([
         node.functionName,
         '(',
-        join(', ', path.map(print, 'arguments')),
+        group(
+          concat([
+            indent(
+              concat([
+                softline,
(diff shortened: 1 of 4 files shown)


src/printer.js
```diff
@@ -499,11 +499,20 @@ function genericPrint(path, options, print) {
       if (node.arguments.length === 0) {
         return node.functionName;
       }
-      // @TODO: add call args
```
fvictorio: @mattiaerre I deleted this comment because it seems outdated, but let me know if there's something pending here.

PR #99, opened 2019-01-07
fix-snap

mattiaerre:
# Description

- fix snap


mattiaerre:
closing because of #98

PR #106, opened 2019-01-11
Fixed tuple errors

maxsam4:
Fixed:
1) Prettier was throwing on `(, yo) = demo(_yo).hello();`
2) `(yo, ) = demo(_yo).hello();` was getting prettified to `(yo) = demo(_yo).hello();`
3) Tupples in VariableExpressions were broken. For example,
`var (,c) = (10, 20);` was getting prettified to `var (, , c) = (10, 20);`
3 was part of the tests but the test screenshot for 3 had the incorrect result so the tests were passing.

fixes #79

PR #107, opened 2019-01-11
Fixed tuples and updated parser

maxsam4:
Tupples in VariableExpressions were broken. For example,
`var (,c) = (10, 20);`
was getting prettified to
`var (, , c) = (10, 20);`

The test screenshot had the incorrect result so the tests were passing.


maxsam4:
Closing in favour of https://github.com/prettier-solidity/prettier-plugin-solidity/pull/106

PR #66, opened 2018-10-16
divide-and-conquer

mattiaerre:
# Description

- separate node types into files

- fix #63 

## (very nice) side effect

```bash
----------------------------------|----------|----------|----------|----------|-------------------|
File                              |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------------------------------|----------|----------|----------|----------|-------------------|
All files                         |    85.52 |     80.5 |     87.5 |    85.52 |                   |
 src                              |    83.71 |    81.63 |     87.5 |    83.71 |                   |
  clean.js                        |       50 |      100 |        0 |       50 |               3,4 |
  index.js                        |    93.33 |    83.33 |      100 |    93.33 |                36 |
  loc.js                          |    88.89 |       75 |      100 |    88.89 |                10 |
  parser.js                       |      100 |      100 |      100 |      100 |                   |
  printer.js                      |    82.52 |    81.75 |      100 |    82.52 |... 03,405,407,425 |
 src/nodes                        |    93.02 |    66.67 |     87.5 |    93.02 |                   |
  ContractDefinition.js           |      100 |      100 |      100 |      100 |                   |
  ImportDirective.js              |       70 |    33.33 |       50 |       70 |          14,16,20 |
  PragmaDirective.js              |      100 |      100 |      100 |      100 |                   |
  SourceUnit.js                   |      100 |      100 |      100 |      100 |                   |
  index.js                        |      100 |      100 |      100 |      100 |                   |
  print-preserving-empty-lines.js |      100 |      100 |      100 |      100 |                   |
----------------------------------|----------|----------|----------|----------|-------------------|
```

PR #121, opened 2019-03-19
Manageable printing Rules

Janther:
The logic of each rule has remained unchanged.
It was only split into smaller and more manageable files.

Easier to write tests and also spot fixes.

Comments and TODOs were also moved into each file


Janther:
I just saw #66 🤣
I believe though this PR takes care of every rule leaving printer.js pretty much like a hub function just making the desition where to forward the message.

PR #122, opened 2019-03-20
4 spaces indentation

Janther:
Solidity industry indents using 4 spaces and this opinionated code formatter should follow the industry.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -55,8 +55,15 @@ const printers = {
   }
 };
 
+// https://prettier.io/docs/en/plugins.html#defaultoptions
+const defaultOptions = {
+  // @TODO fix indentation in block comments
+  tabWidth: 4
+};
+
 module.exports = {
   languages,
   parsers,
-  printers
+  printers,
+  defaultOptions
 };
--- a/tests/AddressPayable/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/AddressPayable/__snapshots__/jsfmt.spec.js.snap
@@ -14,14 +14,14 @@ contract AddressPayable {
 pragma solidity ^0.5.2;
 
 contract AddressPayable {
-  address payable[] hello;
-  function sendSomeEth(address payable to, address payable[] memory world)
-    public
-    payable
-  {
-    address payable target = to;
-    target.transfer(msg.value);
-  }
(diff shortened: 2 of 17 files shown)


mattiaerre:
I've got no particular preference but I see there are some strange indentations in the snapshots?

Janther:
everything that is code related has 4 spaces indentation (I changed these manually before running the tests).
However, block comments look a bit weird. I added a TODO [line](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/122/files#diff-1fdf421c05c1140f6d71444ea2b27638R60)

Also, snapshots contain the code before and after the format so you might also be looking at the code before the format, which contains different types of indentation (2 spaces, 4 spaces, tabs, etc) and trailing whitespace.

fvictorio:
LGTM. @Janther, can you give an example of some block comment that looks weird after being formatted?

Janther:
For example, this is the block before formatting:

https://github.com/prettier-solidity/prettier-plugin-solidity/blob/f80a9a811b53c95e8a40ecb81e5606ea7abb3b7f/tests/BasicIterator/__snapshots__/jsfmt.spec.js.snap#L36-L38

and after formatting it looks like this:

https://github.com/prettier-solidity/prettier-plugin-solidity/blob/f80a9a811b53c95e8a40ecb81e5606ea7abb3b7f/tests/BasicIterator/__snapshots__/jsfmt.spec.js.snap#L80-L82

also tabs are kept within these blocks

https://github.com/prettier-solidity/prettier-plugin-solidity/blob/f80a9a811b53c95e8a40ecb81e5606ea7abb3b7f/tests/BasicIterator/__snapshots__/jsfmt.spec.js.snap#L49-L51

Janther:
trying these out with prettier on javascript, it has the same issue.
I think the Prettier team doesn't mind.

fvictorio:
Yeah, the formatting of comments is handled by prettier itself. If the same thing happens in js, I think we can ignore it here.

Keeping the tabs seems OK to me: it makes sense to keep the comment's content verbatim.

The first example does look wrong though. Maybe you can open an issue in Prettier's repo?

mattiaerre:
if we are all happy I'll merge and deploy

Issue #116, opened 2019-01-23
Misaligned returns when parameters are split


fvictorio:
This

```solidity
contract Foo {
  function foo() modifier1 modifier2 aLongModifier anotherVeryLongModifier returns (uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName) {
    a = 42;
  }
}
```

is prettified as:

```solidity
contract Foo {
  function foo()
    modifier1
    modifier2
    aLongModifier
    anotherVeryLongModifier
    returns (
    uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName
  )
  {
    a = 42;
  }
}
```

Two things are wrong here: the indentation of the returns parameter and closing parentheses, and that the closing parentheses and the opening body brace are not in the same line.

Janther:
I've been working on this for a while and I'd like to talk about the position of the opening body brace.

This is how it is now:
```Solidity
contract Foo {
    function foo()
        modifier1
        modifier2
        aLongModifier
        anotherVeryLongModifier
        returns (
        uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName
    )
    {
        a = 42;
    }
}
```
Fixing the indentation it looks like this:

```Solidity
contract Foo {
    function foo()
        modifier1
        modifier2
        aLongModifier
        anotherVeryLongModifier
        returns (
            uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName
        )
    {
        a = 42;
    }
}
```
Now. if we want the opening body in the same line as the closing parenthesys, it will look like this
```Solidity
contract Foo {
    function foo()
        modifier1
        modifier2
        aLongModifier
        anotherVeryLongModifier
        returns (
            uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName
        ) {
        a = 42;
    }
}
```
now there is no clear separation between the body and the function signature.

there is another option to have the return statement unindented like this:

```Solidity
contract Foo {
    function foo()
        modifier1
        modifier2
        aLongModifier
        anotherVeryLongModifier
    returns (
        uint aVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongParameterName
    ) {
        a = 42;
    }
}
```

[...]

PR #125, opened 2019-03-24
Using semver rules to properly format pragma versions

### LINKED ISSUE #119: pragma with two ranges is not formatted correctly
A pragma like this:

```
pragma solidity >=0.4.21 <0.6.0;
```

gets formatted like this:

```
pragma solidity >=0.4.21<0.6.0;
```

Notice the missing space.

Janther:
Addressing #119 

The issue comes from the way [solidity-parser-antlr](https://github.com/federicobond/solidity-parser-antlr) parses the value of the pragma version.

The package [semver](https://docs.npmjs.com/misc/semver) will format `^1.2.4` as `>=1.2.4 <2.0.0` so I'm applying the function only when more than one range is given to the pragma version.

closes #119


Janther:
I'm using node version `v11.12.0` I'll try with another regex.

mattiaerre:
hi @Janther we are now targeting v6 v8 and v10, we can also drop the support for node v6 (end of life is 2019-04-30) if that helps

Janther:
I'll use a more compatible strategy but will look like a hack (which it is since it's not expected of the value to remove spaces between ranges)

Janther:
Also here is the issue in solidity-parser-antlr
https://github.com/federicobond/solidity-parser-antlr/issues/15

mattiaerre:
@Janther is it something we are able to fix?

src/nodes/PragmaDirective.js
```diff
@@ -4,8 +4,21 @@ const {
...
-  print: ({ node }) => concat(['pragma ', node.name, ' ', node.value, ';'])
+  print: ({ node }) => {
+    // @TODO: remove hack once solidity-parser-antlr is fixed
+    let value = node.value
+      .replace(/([<>=])/g, ' $1')
+      .replace(/< =/g, '<=')
+      .replace(/> =/g, '>=')
+      .trim();
+    if (value.split(' ').length > 1) {
+      value = semver.validRange(value) || node.value;
```
mattiaerre: not sure if we need to add `|| node.value`; can you add a test that uses that branch?

Janther: `semver.validRange()` returns `null` if the string given is not a valid semver range.

In this scenario, I thought it would be better to keep the original `node.value` instead of replacing this with an empty string.

I couldn't trigger this since [solidity-parser-antlr](https://github.com/federicobond/solidity-parser-antlr) takes care of throwing an exception, in case there is an unexpected value. But they already are making a mistake by stripping the spaces between ranges, so who knows.

I added a test for this use case.

mattiaerre: oh, from a test point of view I was thinking more about adding an additional case to the existing `tests/Pragma/Pragma.sol` file

mattiaerre: @Janther let me know if you want me to fix the tests

Janther: Hi @mattiaerre,

I haven't forgotten I just couldn't figure out a scenario where [solidity-parser-antlr](https://github.com/federicobond/solidity-parser-antlr) would return an invalid value for range (other than the one that we are fixing in this PR).

That's why instead of an integration test, I provided a unit test in case we decide to go with a different parser or the current parser changes and allows for an invalid version range (not to mention that there might be a combination that I just missed).

Feel free to add tests and use cases here.

Anyway, this is just meant to be a patch until the parser is fixed.
I've been focusing my effort in #116 at the moment. I'm hoping to show some progress by next week.

mattiaerre: awesome, let's divide and conquer; I'll fix the tests so that you can focus on #116 which I think it is much more important. thanks again @Janther 👍

mattiaerre: it seems to me that it's practically impossible to cover this branch:

Screen Shot 2019-03-28 at 4 23 11 PM

I'd be inclined to remove `|| node.value` if you don't mind @Janther

Janther: I'm ok with this :)

mattiaerre: I'll remove it and merge this PR 👍

PR #135, opened 2019-06-06
Adding a test for clean.js so code coverage will shut up about it

Janther:
Does exactly what it says on the tin.

### DIFF
--- a/__tests__/clean.test.js
+++ b/__tests__/clean.test.js
@@ -1,6 +1,6 @@
 const clean = require('../src/clean');
 
-test('nodes list to match snapshot', () => {
+test('clean removes the unused properties, ', () => {
   let toCleanse = {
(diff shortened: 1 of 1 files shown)


mattiaerre:
this shouldn't be necessary if you run the tests w/ `AST_COMPARE=true npm test`
could you please have a look and confirm?

Janther:
can we configure the Codecov Report using `AST_COMPARE=true` then?
Every PR that I make has a red square in the corner 😂 
Screenshot 2019-06-11 22 45 13

fvictorio:
Alternatively, just ignore that file in the coverage?

mattiaerre:
let me have a look into this

mattiaerre:
shall we close this one because of #151 ?

Janther:
happily

Issue #115, opened 2019-01-23
Don't split function call without arguments


fvictorio:
Running prettier on this contract with a `printWidth` of 56:

```solidity
contract Foo {
  function foo() {
    address validatorAddress = signature.popLast20Bytes();
  }
}
```

outputs this:

```solidity
contract Foo {
  function foo() {
    address validatorAddress = signature.popLast20Bytes(

    );
  }
}
```

I think this shouldn't break like this. In this particular case, I think a proper output might be:

```solidity
contract Foo {
  function foo() {
    address validatorAddress =
      signature.popLast20Bytes();
  }
}
```

or

```solidity
contract Foo {
  function foo() {
    address validatorAddress = signature
      .popLast20Bytes();
  }
}
```

(I prefer the first one and I *think* it's what prettier-js does.)

mattiaerre:
is this still an open issue? @fvictorio @Janther

Janther:
I've got a PR in the oven that will kill these issue

PR #136, opened 2019-06-06
avoid splitting parentheses of a function call without arguments

Janther:
halfway fixing #115
still working on splitting function calls on the `.` character.

### DIFF
--- a/src/nodes/FunctionCall.js
+++ b/src/nodes/FunctionCall.js
@@ -1,48 +1,61 @@
+/* eslint-disable implicit-arrow-linebreak */
 const {
   doc: {
     builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier');
 
-const FunctionCall = {
-  print: ({ node, path, print }) => {
-    let doc;
-    if (node.names && node.names.length > 0) {
-      doc = concat([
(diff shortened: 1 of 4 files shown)


src/nodes/FunctionCall.js
```diff
@@ -1,48 +1,61 @@
...
+      ),
+      softline
+    ])
+  );
+
+const printArguments = (node, path, print) => {
+  if (node.names && node.names.length > 0) {
+    return printObject(node, path, print);
   }
+  if (node.arguments && node.arguments.length > 0) {
```
fvictorio: This doesn't come from this PR, but I'm curious. When does a function call have "names" and when it has "arguments"?

Janther: Will have to double check this one but from the look of the output code, a function call has names when is initializing a struct.
```Solidity
struct FooStruct {
  uint foo;
  uint figther;
}

FooStruct myStruct = FooStruct({foo:1, fighter:2});
```

fvictorio: Oh, I see, thanks!

mattiaerre:
you can rebase now

PR #144, opened 2019-06-07
Respect single quote

### LINKED ISSUE #114: `singleQuote` is not respected
As discussed int https://github.com/0xProject/0x-monorepo/pull/1495, it seems like we are always using double quotes.

In which other places besides string literals should we use this option? Import directives?

fvictorio:
Closes #114.

### DIFF
--- a/src/prettier-comments/common/util.js
+++ b/src/prettier-comments/common/util.js
@@ -430,11 +430,7 @@ function getIndentSize(value, tabWidth) {
   );
 }
 
-function printString(raw, options, isDirectiveLiteral) {
-  // `rawContent` is the string exactly like it appeared in the input source
-  // code, without its enclosing quotes.
-  const rawContent = raw.slice(1, -1);
-
+function printString(rawContent, options) {
   const double = { quote: '"', regex: /"/g };
   const single = { quote: "'", regex: /'/g };
 
@@ -467,17 +463,6 @@ function printString(raw, options, isDirectiveLiteral) {
       : shouldUseAlternateQuote
         ? alternate.quote
         : preferred.quote;
-
-  // Directives are exact code unit sequences, which means that you can't
-  // change the escape sequences they use.
-  // See https://github.com/prettier/prettier/issues/1555
-  // and https://tc39.github.io/ecma262/#directive-prologue
-  if (isDirectiveLiteral) {
-    if (canChangeDirectiveQuotes) {
-      return enclosingQuote + rawContent + enclosingQuote;
-    }
-    return raw;
-  }
 
   // It might sound unnecessary to use `makeString` even if the string already
   // is enclosed with `enclosingQuote`, but it isn't. The string could contain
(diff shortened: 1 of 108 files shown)


src/prettier-comments/common/util.js
```diff
@@ -430,11 +430,7 @@ function getIndentSize(value, tabWidth) {
   );
 }
 
-function printString(raw, options, isDirectiveLiteral) {
-  // `rawContent` is the string exactly like it appeared in the input source
-  // code, without its enclosing quotes.
-  const rawContent = raw.slice(1, -1);
-
+function printString(rawContent, options) {
```
fvictorio: I modified this function to make it more useful for us. It wasn't being used anywhere, so it shouldn't be a problem.

Janther:
any idea why codecov passed and travis failed??

fvictorio:
I have no idea. Tests pass for me locally. Looking at the logs, it seems that jest is having a different behavior when comparing escaped characters. But we have a `package-lock.json` file, so this shouldn't happen. I'll see what I can do.

fvictorio:
@Janther @mattiaerre could you please check if you get an error when running the tests in this branch?

Janther:
When I run the tests with `AST_COMPARE = false` they pass.
When I run the tests with `AST_COMPARE = true` they fail.

mattiaerre:
yes the problem is when you run `AST_COMPARE=true npm t`; you get the errors you see in CI and for sure the problem is in `src/prettier-comments/common/util.js`

fvictorio:
I see, thanks! I'll take a look.

fvictorio:
I have the feeling that this is an upstream problem. I opened an issue in the parser's repo: https://github.com/federicobond/solidity-parser-antlr/issues/66

mattiaerre:
@fvictorio I agree w/ you actually; hopefully @federicobond can help us here

fvictorio:
He confirmed it's a bug. I'll try to send him a PR as soon as I can.

mattiaerre:
this is awesome, thanks for that @fvictorio

fvictorio:
I upgraded `solidity-parser-antlr` and now the problem with the escaped quotes is fixed. But now there was an issue with one of the tests. Before this, if a modifer had parentheses but no arguments, the parentheses were removed. For example, `function foo() myModifier()` would become `function foo() myModifier`. I actually like this, but now the parser seems to distinguish between the two cases. I changed it so that the parentheses are kept if they were present, and they are not added if they weren't.

If we want the previous behavior (removing parentheses when a modifier has no arguments), we'll need to modifier either the parser or the AST check.

Thoughts? @mattiaerre @Janther

fvictorio:
Oh, disregard that, I merged master into this branch and that change was already there, so I guess everything is fine.

fvictorio:
The CI was failing only for node 8 because `npm ls` was exiting with 1. I tried it locally with the same npm version and I could reproduce it. I then upgraded npm and it didn't happen. So I added a `before_install` script to Travis to use a specific npm version.

mattiaerre:
this is awesome, thanks so much @fvictorio

Issue #145, opened 2019-06-07
long arrays not breaking down correctly


codfish:
Long arrays don't seem to be breaking down correctly. Using solium and it complains about indentation post-prettier:

For example, I expect the following:
```c
bytes32[] public permissions = [bytes32("addPartner"), bytes32("removePartner"), bytes32("addGroup"), bytes32("removeGroup"), bytes32("addUserToGroup"), bytes32("removeUserFromGroup"), bytes32("addSolution"), bytes32("removeSolution"), bytes32("addLibrary"), bytes32("removeLibrary"), bytes32("addPermissionToGroup"), bytes32("removePermissionFromGroup")];
```

would format to:

```c
bytes32[] public permissions = [
    bytes32("addPartner"),
    bytes32("removePartner"),
    bytes32("addGroup"),
    bytes32("removeGroup"),
    bytes32("addUserToGroup"),
    bytes32("removeUserFromGroup"),
    bytes32("addSolution"),
    bytes32("removeSolution"),
    bytes32("addLibrary"),
    bytes32("removeLibrary"),
    bytes32("addPermissionToGroup"),
    bytes32("removePermissionFromGroup")
];
```

but it formats incorrectly to:
```c
bytes32[] public permissions = [bytes32("addPartner"), bytes32("removePartner"), bytes32(
        "addGroup"
    ), bytes32("removeGroup"), bytes32("addUserToGroup"), bytes32("removeUserFromGroup"), bytes32(
        "addSolution"
    ), bytes32("removeSolution"), bytes32("addLibrary"), bytes32("removeLibrary"), bytes32(
        "addPermissionToGroup"
    ), bytes32("removePermissionFromGroup")];
```

and solium fails with a bunch of `indentation` failures

Janther:
Wow I love this one.
I'll see what I can do here.

mattiaerre:
closing it thanks to #146

Issue #140, opened 2019-06-06
a long return statement should be split


Janther:
Currently this:
```Solidity
function thisFunctionNameIsReallyLong()
    public
    returns (
        address someAddressName,
        uint256 LongArgument,
        uint256 Argument
    )
{
    doSomething();

    return (veryVeryLongReturnArg1,
            veryVeryLongReturnArg1,
            veryVeryLongReturnArg1);
}
```
will be prettyfied like this:
```Solidity
function thisFunctionNameIsReallyLong()
    public
    returns (
        address someAddressName,
        uint256 LongArgument,
        uint256 Argument
    )
{
    doSomething();

    return (veryVeryLongReturnArg1, veryVeryLongReturnArg1, veryVeryLongReturnArg1);
}
```
Any comments @fvictorio @mattiaerre ?

fvictorio:
It makes sense, and it might be a good first issue.

minggas:
I can give i try with this. The file that need to be modified is `prettier-plugin-solidity/src/nodes/ReturnStatement.js`?

Janther:
Yes and also add tests around this particular use :)

codfish:
This might be typical prettier behavior unless that line is longer than `max-len`. At least in JS this would be normal.

```js
return { veryVeryLongReturnArg1, veryVeryLongReturnArg1, veryVeryLongReturnArg1 };
```
^ would stay like this if it's under the `printWidth`

Janther:
Hi @codfish that's what I meant by "a long return statement". Thank you for clarifying.

```Solidity
// return statment shorter than printWith
return (ReturnArg1, ReturnArg1, ReturnArg1);

// return statment longer than printWith
return (
    veryVeryLongReturnArg1,
    veryVeryLongReturnArg1,
    veryVeryLongReturnArg1
);
```

mattiaerre:
closing it thanks to #146

PR #146, opened 2019-06-08
fixing array indentation and also solved the return statements.

Janther:
Solving #145, by chance this also solves #140 (sorry @minggas).

However this uncovered 2 hidden bugs.

1. Trailing comments are considered in the `printWidth` rule.
2. `if` and `while` and `for` conditions should break in the logic operators. `||`, `&&`, `<`, `>`,`<=`, `>=`, `==`

### DIFF
--- a/src/nodes/TupleExpression.js
+++ b/src/nodes/TupleExpression.js
@@ -1,23 +1,29 @@
+/* eslint-disable implicit-arrow-linebreak */
 const {
   doc: {
-    builders: { concat, join }
+    builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier');
 
 const TupleExpression = {
-  print: ({ node, path, print }) => {
-    let doc;
-    // @TODO: remove hack once solidity-parser-antlr is fixed
-    if (node.components) {
-      doc = join(', ', path.map(print, 'components'));
-    } else {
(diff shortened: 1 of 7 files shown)


fvictorio:
I don't get why a change in `TupleExpression` affects condition expressions inside `if`s :confused:

Janther:
is not the `if` condition that gets broken, is the `TupleExpression` within the `if`

```Solidity
// This is ok but not great
if (veryLongConditionWontBreak) {}

// It should look like this
if (
    veryLongConditionShouldBreak
) {}

// This looks weird
if ((
    4 + 5
) / 6 < thisIsAVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName) {
}

// It should look like this
if (
    (4 + 5) / 6 <
    thisIsAVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryVeryLongName
) {
}
```
But this is a bug that will be solved as we define better grouping within each node.

fvictorio:
Oh, I see. Weird that those kind of expressions are called `TupleExpressions`.

Janther:
I believe Tuple refers to the pair of parentheses `()` or `[]`, so these apply to arrays, returns and parentheses in any computation.

PR #147, opened 2019-06-10
Feature, browser friendly

Janther:
I'm experimenting using prettier within the browser and was surprised by all of the warnings and difficulty to run.

Then I noticed prettier also provides a standalone script for browsers.

This script mainly doesn't load plugins from the filesystem and you have to provide the plugin in a javascript object.

By requiring `prettier` we are loading also all of the filesystem's libraries into the build (probably Webpack then removes them as we don't use them) but warnings are raised and also a feeling of uneasiness.

This Pull Request uses `prettier/standalone` to load all of the core builders.
Also (and I'm very sorry for this) I had to remove `dir-to-object` from this plugin's dependencies and explicitly require each File. This could be done in a `prepublish` script but I feel like we would be adding an extra step just to avoid having an explicit list of files.

Now, this plugin is fully compatible with the browser and the node environments.


Janther:
note that the tests are still run against `prettier` and not `prettier/standalone` thus ensuring that the main functionality is untouched.

mattiaerre:
uhm @Janther this is really interesting; let me have a better look at this

mattiaerre:
uhm ... this has not been merged for sure 🤔

Janther:
Noo I pressed the wrong button. I was trying to close this. Then I undid the changes and made the PR again

Issue #130, opened 2019-06-04
.eslintrc


mattiaerre:
# Description

- add `implicit-arrow-linebreak`

fvictorio:
This seems like the kind of eslint rule that you can't use along with prettier (core prettier I mean, the one used here for formatting javascript). In fact, [`eslint-config-prettier`](https://github.com/prettier/eslint-config-prettier/blob/d3a0c0d80ca04ab3f1d106481ea2c3bb39abc04f/index.js#L36) disables it. We aren't using that config right now, but it's very common to include it when using prettier.

minggas:
Hi, can I give a try?

fvictorio:
Hi @minggas! I don't know if we are going to do this, but maybe you want to give https://github.com/prettier-solidity/prettier-plugin-solidity/issues/140 a try? If you need some pointers, we'd be happy to help you.

mattiaerre:
Hi @fvictorio you are right, we are `"extends": "airbnb"` and not using `eslint-config-prettier`; shall we rename this issue in use `eslint-config-prettier` instead? in that case woul @minggas give it a try?

fvictorio:
That sounds like a good idea, yes.

minggas:
I like to give a try if it's ok with you.

The change will be on the .eslintrc:
from:
```
"rules": {
    "arrow-parens": "off",
    "comma-dangle": "off",
    "object-curly-newline": "off"
  }
```
to:
```
"rules": {
    "arrow-parens": "off",
    "comma-dangle": "off",
    "object-curly-newline": "off",
    "eslint-config-prettier": "off"
  }
```
am i right?

fvictorio:
No, that won't work. `eslint-config-prettier` is a config, not a rule. Check the "Installation" section in https://github.com/prettier/eslint-config-prettier docs.

mattiaerre:
I'm gonna close this as #149 has been merged, thanks @minggas

PR #149, opened 2019-06-11
Installing eslint-prettier-config and adding it to .eslintrc

minggas:
Install `eslint-config-prettier` and add `"prettier"` to `"extends"` on `.eslintrc` to solve #130

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,10 @@
   "env": {
     "jest": true
   },
-  "extends": "airbnb",
+  "extends": [
+    "airbnb",
+    "prettier"
+  ],
   "globals": {
     "run_spec": false
   },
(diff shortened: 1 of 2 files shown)


minggas:
What was the problem?

Janther:
Relax @minggas nothing is wrong with this (but I know little about eslint's configurations).
The coverage didn't pass because of `src/clean.js`
We are discussing it in #135

mattiaerre:
you should be able to either rebase or create another pr from master @minggas now that #151 has been merged

.eslintrc
```diff
@@ -2,7 +2,10 @@
   "env": {
     "jest": true
   },
-  "extends": "airbnb",
+  "extends": [
+    "airbnb",
```
mattiaerre: should we be also getting rid of `airbnb`?

fvictorio: I don't think so, it has some useful rules.

fvictorio:
This looks fine. My only comment is that I think now we can remove the rules that we have explicitly disabled in the configuration.

mattiaerre:
I am going to merge this a create a followup PR w/ some more cleanup

PR #156, opened 2019-06-12
Browser friendly

Janther:
I'm experimenting using prettier within the browser and was surprised by all of the warnings and difficulty to run.

Then I noticed prettier also provides a [standalone](https://prettier.io/docs/en/browser.html) script for browsers.

This script mainly doesn't load plugins from the filesystem and you have to provide the plugin in a javascript object.

By requiring `prettier` we are loading also all of the filesystem's libraries into the build (probably Webpack then removes them as we don't use them) but warnings are raised and also a feeling of uneasiness.

This Pull Request uses `prettier/standalone` to load all of the core builders.
Also (and I'm very sorry for this) I had to remove `dir-to-object` from this plugin's dependencies and explicitly require each File. This could be done in a `prepublish` script but I feel like we would be adding an extra step just to avoid having an explicit list of files.

Now, this plugin is fully compatible with the browser and the node environments.


Janther:
@mattiaerre redid the PR now has no red files in the corner :)

src/nodes/index.js
```diff
@@ -1,5 +1,121 @@
...
+const UserDefinedTypeName = require('./UserDefinedTypeName.js');
+const UsingForDeclaration = require('./UsingForDeclaration.js');
+const VariableDeclaration = require('./VariableDeclaration.js');
+const VariableDeclarationStatement = require('./VariableDeclarationStatement.js');
+const WhileStatement = require('./WhileStatement.js');
 
-const nodes = dirToObject(__dirname, { canAdd: data => data.print });
-
-module.exports = nodes;
+module.exports = {
```
mattiaerre: it'd be nice to find a way to dynamically do this; 🤔

src/prettier-comments/language-js/comments.js
```diff
@@ -7,8 +7,8 @@ const {
     addDanglingComment,
     getNextNonSpaceNonCommentCharacterIndex
   }
-} = require('prettier');
-const privateUtil = require("../common/util");
+} = require('prettier/standalone');
+const privateUtil = require('../common/util');
```
mattiaerre: can we keep the `"` in this file as it was just copied pasted from the prettier main repo?

Janther: done

mattiaerre:
hi @Janther what do you think about my last commit? also do you mind to resolve the conflicts? thanks

mattiaerre:
there is also I failing test unfortunately

Janther:
Rebased it and made the necessary changes :)

mattiaerre:
oh noes ... I was still working on this 😭

Janther:
😮

Janther:
I just moved your script to a `scripts` folder and slightly changed it so it would receive any folder

mattiaerre:
I've changed the way we should invoke it and also added some code coverage for it; what about creating another PR?

Janther:
Don't worry, ;) I'll have a look after a meeting I have now.

Janther:
also it looks like we had the same idea

Janther:
fixed.
had to change the `new Date(Date.now()).toString()` and just do `Date.now()` because of me being in a different timezone was making the tests fail

fvictorio:
I was thinking that it would be nice to do a Remix plugin for prettier, so a lot of :+1: to this.

Janther:
I did a remix plugin using this branch :)
https://remix-prettier.netlify.com

fvictorio:
Oh, nice!! We could add it to the readme when this is merged.

Issue #138, opened 2019-06-06
Force braquets when if statement is longer than a line


Janther:
According to the [Style Guide](https://solidity.readthedocs.io/en/v0.5.9/style-guide.html#control-structures) if we have more than a line in a `trueBody` or a `falseBody`, we should enforce braquets.

Before Prettier.
```Solidity
if (x < 10)
    someArray.push(Coin({
        name: 'spam',
        value: 42
    }));
```
After Prettier.
```Solidity
if (x < 10) {
    someArray.push(Coin({
        name: 'spam',
        value: 42
    }));
}
```

fvictorio:
This seems like a non-trivial implementation, but I think it can be done. Maybe it makes sense to fix this and #137 (if it's accepted) together.

Janther:
Theoretically this can be done using the [ifBreak](https://github.com/prettier/prettier/blob/master/commands.md#ifbreak) builder.
```Javascript
concat([ifBreak('{', ''), trueBody, ifBreak('}', '')])
```
I'll have to test it though.

Janther:
ok, I did this one and I found a problem.

If I take this:
```Solidity
if (true) return true;
```

the AST for the trueBody is:
```JSON
"trueBody": {
    "expression": {
      "type": "BooleanLiteral",
      "value": true,
    },
    "type": "ReturnStatement",
  }
}
```

If I take this:
```Solidity
if (true) { return true; }
```

the AST for the trueBody is:
```JSON
"trueBody": {
  "statements": [
    {
      "expression": {
        "type": "BooleanLiteral",
        "value": true,
      },
      "type": "ReturnStatement",
    },
  ],
  "type": "Block",
}
```

Modifying the AST compare test for this use case is something we need to discuss.

This would benefit `if`, `while`, and `for` statements.
The changes are in this [branch](https://github.com/prettier-solidity/prettier-plugin-solidity/tree/feature/add_brackets_if_while_for_body_when_splits)

Please have a look.
@mattiaerre @fvictorio

Janther:
also, this introduces a small regression for the particular case if a single line `trueBody` followed by an `else`.
```Solidity
if (ifCondition) return true;
else return false;

if (ifCondition) return true;
else if (elseIfCondition) return (true, false);
else return false;
```

will be prettified as
```Solidity
if (ifCondition) {
    return true;
} else return false;

if (ifCondition) {
    return true;
} else if (elseIfCondition) {
[...]

PR #159, opened 2019-06-13
Indent if while for body when splits

Janther:
while solving #138
I found that there's a mismatch between the AST for `if (x) return true;` and `if (x) { return true; }`.
Nonetheless, I found some benefits in the exercise of creating the feature so for the moment as a compromise, this PR indents the bodies of an `IfStatement`, `ForStatement`, and `WhileStatement` in case of them being too long.
I also added the case of forcing the indentation of a nested `if`.
```Solidity
// Before Prettier
if (a) if (b) return true; else return false; else if (c) break; else throw;

// After Prettier
if (a)
    if (b) return true;
    else return false;
else if (c) break;
else throw;
```

### DIFF
--- a/src/nodes/IfStatement.js
+++ b/src/nodes/IfStatement.js
@@ -4,19 +4,40 @@ const {
   }
 } = require('prettier');
 
-const printBody = (body, followingElse, node, path, print) => {
-  if (node[body].type === 'Block' || node[body].type === 'IfStatement') {
-    return concat([' ', path.call(print, body), followingElse ? ' ' : '']);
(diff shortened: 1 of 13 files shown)


src/nodes/IfStatement.js
```diff
@@ -1,38 +1,58 @@
 /* eslint-disable implicit-arrow-linebreak */
```
mattiaerre: we shouldn't have these lines anymore

Janther: I was waiting for #157 to be merged. now I rebased to master.

PR #164, opened 2019-06-20
indenting chained member and function call

### LINKED ISSUE #115: Don't split function call without arguments

Janther:
closes #115 
Before this PR

```Solidity
veryLongVariable.veryLongMember.veryLongMember.veryLongMember.veryLongMember.veryLongMember;
```

After this PR
```Solidity
veryLongVariable
    .veryLongMember
    .veryLongMember
    .veryLongMember
    .veryLongMember
    .veryLongMember;
```

This cleans up the code in many places, so the snapshot where updated and reviewed.

### DIFF
--- a/src/nodes/MemberAccess.js
+++ b/src/nodes/MemberAccess.js
@@ -1,12 +1,30 @@
 const {
   doc: {
-    builders: { concat }
+    builders: { concat, group, indent, softline }
   }
 } = require('prettier');
 
+const isBeginnigOfChain = path => {
+  const parentNodeType = path.getParentNode().type;
+
+  if (parentNodeType === 'MemberAccess') return false;
+  if (parentNodeType === 'FunctionCall') {
+    const grandParentNodeType = path.getParentNode(1).type;
+    return grandParentNodeType !== 'MemberAccess';
+  }
+
+  return true;
+};
+
 const MemberAccess = {
-  print: ({ node, path, print }) =>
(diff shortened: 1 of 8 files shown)


src/nodes/MemberAccess.js
```diff
@@ -1,12 +1,30 @@
...
   }
 } = require('prettier');
 
+const isBeginnigOfChain = path => {
+  const parentNodeType = path.getParentNode().type;
+
+  if (parentNodeType === 'MemberAccess') return false;
+  if (parentNodeType === 'FunctionCall') {
+    const grandParentNodeType = path.getParentNode(1).type;
+    return grandParentNodeType !== 'MemberAccess';
```
mattiaerre: 😅 love this name

PR #167, opened 2019-06-26
 Adding support for the common configuration spacingBrackets

Janther:
Additionally added the options object.
Prettier would not accept the default option `false` until I described it in this object.
Reading the official plugins, they relist every supported option, so I figured we do the same here.

As following the style guide, the default is no space.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,0 +1,54 @@
+const CATEGORY_GLOBAL = 'Global';
+const CATEGORY_COMMON = 'Common';
+
+const options = {
+  bracketSpacing: {
+    since: '0.0.0',
+    category: CATEGORY_COMMON,
+    type: 'boolean',
+    default: false,
+    description: 'Print spaces between brackets.',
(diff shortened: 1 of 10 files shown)


src/options.js
```diff
@@ -0,0 +1,54 @@
...
+  useTabs: {
+    since: '1.0.0',
+    category: CATEGORY_GLOBAL,
+    type: 'boolean',
+    default: false,
+    description: 'Indent with tabs instead of spaces.'
+  }
+};
+
+module.exports = options;
```
mattiaerre: noice this is really cool @Janther 👍

tests/EnumDefinitions/jsfmt.spec.js
```diff
@@ -0,0 +1,2 @@
+run_spec(__dirname);
+run_spec(__dirname, { bracketSpacing: true });
```
mattiaerre: I like this, very good

PR #169, opened 2019-06-28
update-deps

mattiaerre:
# Description

- update dependencies

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
   "homepage": "https://github.com/prettier-solidity/prettier-plugin-solidity#readme",
   "devDependencies": {
     "codecov": "^3.0.4",
-    "eslint": "^6.0.1",
+    "eslint": "^5.16.0",
     "eslint-config-airbnb": "^17.0.0",
     "eslint-config-prettier": "^6.0.0",
     "eslint-plugin-import": "^2.18.0",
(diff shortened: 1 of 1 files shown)


Janther:
I downgraded the versions of these packages to support `node 6`

| Package                | Current | Wanted | Latest |
| ---------------------- | ------- | ------ | ------ |
| `escape-string-regexp` | 1.0.5   | 1.0.5  | 2.0.0  |
| `eslint`               | 5.16.0  | 5.16.0 | 6.0.1  |
| `string-width`         | 3.1.0   | 3.1.0  | 4.1.0  |

If we want to upgrade these we will have to stop supporting `node 6` in our travis config.

mattiaerre:
@Janther yes I've noticed that that was the problem, I'm down to stop supporting node 6 tbh, what do you think?

Janther:
It’s very tricky to decide which versions to use. In my own projects I tend
to use always the latest versions of packages but in an open source one we
might want to keep using older ones as long as we don’t find a critical
issue.
Anyway at the moment we don’t have much info on how much this tool is
being used and in which scenarios.
I recon it’s used in new versions of node only.
I would add an `engines` attribute in our `package.json` if we are not
gonna support a node version.

mattiaerre:
Yeah, I agree. let's do this 👍

PR #171, opened 2019-06-28
Adding the option to standardise the type declarations

Janther:
This PR will enable the team to decide what type of unit declaration they would rather have across the project.
`explicitTypes: always` will force every instance of `uint`, `int`, and `byte` to be expanded as `uint256`, `int256`, and `bytes1`.
The option `explicitTypes: never` will prefer `uint`, `int`, and `byte` over the aliases.
Finally the option `explicitTypes: preserve` will leave the declaration untouched.

The manipulation is done in the parsing step, so there will be no problem when comparing the AST.

As a best practice, I choose to have this option as `always` by default, but we can change this.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -1,5 +1,6 @@
 const CATEGORY_GLOBAL = 'Global';
 const CATEGORY_COMMON = 'Common';
+const CATEGORY_SOLIDITY = 'Solidity';
 
 const options = {
   bracketSpacing: {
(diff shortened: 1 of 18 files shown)


mattiaerre:
hi @Janther I've rebased from master, let me know if I've done something wrong please

Janther:
it was a merge but LGTM. once you merge, it will flatten all the commits into 1

PR #172, opened 2019-07-01
Contract Definition comments

### LINKED ISSUE #162: Empty contract won't print comments inside.
This contract:

```Solidity
pragma solidity ^0.5.0;

contract Comments {
    // solhint-disable-previous-line no-empty-blocks
}
```

throws the following error.
```
Comment "solhint-disable-previous-line no-empty-blocks" was not printed. Please report this error!
```

Janther:
fixes #162 

in this PR there are 2 new features.

  1. the indentation on `BlockComment`s is now aligned with the indentation of the subject of that comment.
  2. many edge cases on comments around contract definitions are addressed.

input
```Solidity
contract A {
    // solhint-disable-previous-line no-empty-blocks
}
```

output won't even print currently.

after this PR
```Solidity
contract A {
    // solhint-disable-previous-line no-empty-blocks
}
```

Scenarios, where we were depending on prettier to guess where the comment should go, are now addressed.
input:

```Solidity
contract A /*comment about the name*/ is X /*interface X because ...*/, Y /*contract Y because ...*/, Z /*implements Z because ...*/ {
    // solhint-disable-previous-line no-empty-blocks
}
```
Output:
```Solidity
contract A is
    /*comment about the name*/
    X, /*interface X because ...*/
    Y, /*contract Y because ...*/
    Z /*implements Z because ...*/
    // solhint-disable-previous-line no-empty-blocks
{}
```

After this PR:
```Solidity
/*comment about the name*/
contract A is
    X, /*interface X because ...*/
    Y, /*contract Y because ...*/
    Z /*implements Z because ...*/
{
    // solhint-disable-previous-line no-empty-blocks
}
```

To address the Contract Definition edge cases I created the following functions
`solidityHandleOwnLineComment`, `solidityHandleEndOfLineComment`, and `solidityHandleRemainingComment`, these functions try to match the comments with Solidity's cases first and if no match was found, they call the javascript comment handler respectively.

My idea is to bring add Solidity's edge cases in this folder. and even redoing the javascript handlers in a way that makes sense with Solidity's AST.
With time we will rely less on the javascript `handlers` while fully using the `utils`.

Finally I had to add the `package-lock.json` because Travis kept installing the wrong eslint and failing to run the linting.

### DIFF
--- a/src/clean.js
+++ b/src/clean.js
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-unused-vars
 function clean(ast, newObj, parent) {
-  ['code', 'codeStart', 'loc', 'range'].forEach(name => {
+  ['code', 'codeStart', 'loc', 'range', 'raw'].forEach(name => {
     delete newObj[name]; // eslint-disable-line no-param-reassign
   });
 }
--- a/src/index.js
+++ b/src/index.js
@@ -1,11 +1,12 @@
-const { handleComments } = require('./prettier-comments');
+// const { handleComments } = require('./prettier-comments');
+// const printComment = require('./comments/printer');
+const { handleComments, printComment } = require('./comments');
 
 const massageAstNode = require('./clean');
 const loc = require('./loc');
 const options = require('./options');
 const parse = require('./parser');
 const print = require('./printer');
-const printComment = require('./print_comment');
 
 // https://prettier.io/docs/en/plugins.html#languages
 const languages = [
--- a/src/index.js
+++ b/src/index.js
@@ -5,6 +5,7 @@ const loc = require('./loc');
 const options = require('./options');
 const parse = require('./parser');
 const print = require('./printer');
+const printComment = require('./print_comment');
 
 // https://prettier.io/docs/en/plugins.html#languages
 const languages = [
@@ -22,24 +23,8 @@ const parsers = {
   'solidity-parse': parser
 };
 
-function canAttachComment(node) {
-  return (
-    node.type && node.type !== 'BlockComment' && node.type !== 'LineComment'
-  );
-}
-
-function printComment(commentPath) {
-  const comment = commentPath.getValue();
-  switch (comment.type) {
-    case 'BlockComment': {
-      return `/*${comment.raw}*/`;
-    }
-    case 'LineComment':
-      return `//${comment.raw.trimRight()}`;
-    default:
-      throw new Error(`Not a comment: ${JSON.stringify(comment)}`);
-  }
-}
+const canAttachComment = node =>
+  node.type && node.type !== 'BlockComment' && node.type !== 'LineComment';
 
 // https://prettier.io/docs/en/plugins.html#printers
 const printers = {
--- a/tests/Proxy/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/Proxy/__snapshots__/jsfmt.spec.js.snap
@@ -63,23 +63,23 @@ pragma solidity ^0.4.24;
  */
 contract Proxy {
     /**
-    * @dev Tells the address of the implementation where every call will be delegated.
-    * @return address of the implementation to which it will be delegated
-    */
+     * @dev Tells the address of the implementation where every call will be delegated.
+     * @return address of the implementation to which it will be delegated
+     */
     function _implementation() internal view returns (address);
 
     /**
-    * @dev Fallback function.
-    * Implemented entirely in \`_fallback\`.
-    */
+     * @dev Fallback function.
+     * Implemented entirely in \`_fallback\`.
+     */
     function _fallback() internal {
         _delegate(_implementation());
     }
 
     /**
-    * @dev Fallback function allowing to perform a delegatecall to the given implementation.
-    * This function will return whatever the implementation call returns
-    */
+     * @dev Fallback function allowing to perform a delegatecall to the given implementation.
+     * This function will return whatever the implementation call returns
+     */
     function _delegate(address implementation) internal {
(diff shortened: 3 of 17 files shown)


.gitignore
```diff
@@ -62,6 +62,4 @@ typings/
 
 .DS_Store
 
-package-lock.json
```
mattiaerre: not sure if I like this idea, we may want to instruct travis to use the proper eslint version perhaps

src/clean.js
```diff
@@ -1,6 +1,6 @@
 // eslint-disable-next-line no-unused-vars
 function clean(ast, newObj, parent) {
-  ['code', 'codeStart', 'loc', 'range'].forEach(name => {
+  ['code', 'codeStart', 'loc', 'range', 'raw'].forEach(name => {
```
mattiaerre: even here, not sure if this is how we should be dealing w/ this

src/index.js
```diff
@@ -1,4 +1,6 @@
-const { handleComments } = require('./prettier-comments');
+// const { handleComments } = require('./prettier-comments');
+// const printComment = require('./comments/printer');
+const { handleComments, printComment } = require('./comments');
```
mattiaerre: uhm interesting

tests/Proxy/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -62,23 +62,23 @@ pragma solidity ^0.4.24;
  */
 contract Proxy {
     /**
-    * @dev Tells the address of the implementation where every call will be delegated.
-    * @return address of the implementation to which it will be delegated
-    */
+     * @dev Tells the address of the implementation where every call will be delegated.
```
mattiaerre: I like this 👍

.gitignore
Janther: There's no rush for this PR so I'll see if after a rebase there is some improvement.

src/clean.js
Janther: in my opinion, comments shouldn't even be in AST comparison.
Prettier throws an exception already if we failed to print a comment.

tests/Proxy/__snapshots__/jsfmt.spec.js.snap
Janther: me too. maybe I could split this PR into
1. fixing comment indentation.
2. Contract Definition edge cases

.gitignore
Janther: I rebased the branch to the latest release with the updated dependencies and the travis uses the right `eslint` without needing `package-lock.json`

fvictorio: @mattiaerre I'm curious, why don't you like this idea? AFAIK versioning the `package-lock.json` or `yarn.lock` is a pretty standard practice.

fvictorio:
@Janther What is this PR missing? Can I help somehow?

Janther:
@fvictorio sorry for the wait, actually I'm happy with the PR but I'm open to improvements/suggestions to be made to it.

As stated in #181 the parser will begin including natspec comments from version `0.5.0` and that breaks some things so I believe we should start processing comments in a solidity way instead of a JS way, and I believe this is the first step towards that so I'd like us all to be on board with it.

mattiaerre:
can you resolve the conflicts here too @Janther ? thanks

yxliang01:
Looking forward to this PR being merged! This is really important to my project :) & thanks for your work!

mattiaerre:
I'd be inclined to merge this PR; shall we add more tests around it so that the cod cov will not drop too much?

Janther:
Conflicts resolved, not sure why codecov is still angry, basically the only 2 lines that are not covered are the exception when there's no printer for this node (the parser had an update) or there's no printer for this comment (extract comments had an update).

please have a look @mattiaerre @fvictorio

mattiaerre:
the problem here is that we've removed `raw` from the clean function and so codecov is complaining as it runs tests w/o the `AST_COMPRARE` flag; I'll be happy to go ahead and merge as it is since this is only affecting comments; what do you think @Janther @fvictorio ?

mattiaerre:
FYI:

https://community.codecov.io/t/mismatch-between-project-and-patch-coverage/1059

// cc @Janther @fvictorio

fvictorio:
I'm OK with merging this.

jerrode:
Hey everyone, Jerrod from @codecov here. Love the high coverage / testing approach you've taken.

Quick note: The check that is failing is `patch` coverage not `project` coverage. Patch = the coverage on the incremental branch, which basically means you must hit 100% coverage on added code. 

image

If there is a specific test that isn't covering that you think should be, have you checked the underlying coverage report to see if the line is covered out of your CI?

https://codecov.io/gh/prettier-solidity/prettier-plugin-solidity/commit/a5bb07fcb86ba1c579f44090bb7558d66fbd22a0/build

fvictorio:
Hi @jerrode, thanks for the help!

@Janther @mattiaerre Apparently the uncovered bit is in `printer.js`: the `throw` in the `default` case for the comment type. It sounds like we are just being defensive there, so I think we can just ignore it.

mattiaerre:
thanks so much for pointing that out @jerrode that's awesome. so you are suggesting @fvictorio that we can just merge this as is? I'm fine w/ that

fvictorio:
It's up to @Janther, but yes, I'm ok with merging this.

Janther:
I’m happy to merge, I know what’s not being covered, I just don’t know why codecov is that angry hehe

Janther:
Can’t merge from my phone though

PR #174, opened 2019-07-03
Exponentiation

### LINKED ISSUE #142: whitespace around and operator can be omitted
In the other recommendations section fo the [Style Guide](https://solidity.readthedocs.io/en/v0.5.9/style-guide.html#other-recommendations) it is said that an operator is allowed to have no whitespace around it in order to express the priority of execution.

```Solidity
x = 2**3 + 5;
x = 2*y + 3*z;
x = (a+b) * (a-b);
```

I'm not that fond of this one.
Comments? @fvictorio @mattiaerre

Janther:
fixes #142 

one use case for the binary operations that was left behind is that the exponentiation operation `**` is always grouped in prettier javascript.

Example:
```Javascript
a =
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d *
  c ** d;
```

This is to denote precedence and also a stubbornness to be separated thus showing the developer that this is a unity.
The same can be seen in other languages where the `**` operator doesn't have spaces around it.
Even the style guide allows not having spaces in order to denote precedence as shown on #142.

In this PR we are following these 2 rules.

1. Prettier will try to fit the `**` operation will be tried to be fit in one line always.
2. There won't be spaces around the `**` operator unless the line has to break.

by default, the option will be no spaces (we might be inspired by Prettier JS but not necessarily follow everything).


fvictorio:
My two (bikeshedding) cents: I would remove the option and always do `a**b`. I think it's a reasonable way of formatting it, and IMO we should avoid adding options if possible.

Janther:
option removed :)

PR #178, opened 2019-08-01
Avoid TypeError:

jablko:
Cannot read property 'range' of null

### DIFF
--- a/src/loc.js
+++ b/src/loc.js
@@ -4,7 +4,7 @@ function getRange(index, node) {
   if (node.range) {
     return node.range[index];
   }
-  if (node.expression.range) {
+  if (node.expression && node.expression.range) {
     return node.expression.range[index];
   }
   return null;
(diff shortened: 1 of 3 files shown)


mattiaerre:
thanks for this PR @jablko do you think you are able to add a test that covers this branch?

jablko:
> thanks for this PR @jablko do you think you are able to add a test that covers this branch?

Done: I've now added an input that triggers this error.

fvictorio:
Unrelated to this PR, but maybe we should format empty fors as `for (;;)`?

mattiaerre:
that's a good point @fvictorio , what do you think we should be doing? shall I wait to merge this PR?

 // cc @jablko @Janther

fvictorio:
I'll just open an issue, I think we can merge this.

Janther:
:thumbsup:

yxliang01:
I'm thinking whether this PR will create silent errors. I feel like prettier-plugin-solidity is in fact expecting `node.expression` is defined at the time that if-condition is being executed. As prettier-plugin-solidity has been working okay, I suspect that only when the AST is somehow unexpected will cause `TypeError` in this case (e.g. breaking changes of `solidity-parser-antlr`). As such, maybe an `assert` statement would be more appropriate?

mattiaerre:
thanks for this feedback @yxliang01 ; what do you think about this @Janther @fvictorio ?
// cc @jablko

PR #180, opened 2019-08-03
Avoid having small right operand on its own line

jablko:
Would you consider grouping the right operand, like [language-js](https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L5899), "to avoid having a small right part like -1 be on its own line"?

### DIFF
--- a/src/binary-operator-printers/comparison.js
+++ b/src/binary-operator-printers/comparison.js
@@ -23,13 +23,17 @@ module.exports = {
   print: (node, path, print) => {
     const indentIfNecessary = indentIfNecessaryBuilder(path);
 
+    const right = concat([node.operator, line, path.call(print, 'right')]);
+    // If it's a single binary operation, avoid having a small right
+    // operand like - 1 on its own line
+    const shouldGroup =
+      node.left.type !== 'BinaryOperation' &&
+      path.getParentNode().type !== 'BinaryOperation';
     return group(
       concat([
         path.call(print, 'left'),
         ' ',
-        indentIfNecessary(
-          concat([node.operator, line, path.call(print, 'right')])
-        )
+        indentIfNecessary(shouldGroup ? group(right) : right)
       ])
     );
   }
--- a/tests/BinaryOperators/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/BinaryOperators/__snapshots__/jsfmt.spec.js.snap
@@ -246,8 +246,7 @@ contract ArithmeticOperators {
             veryVeryVeryVeryVeryLongVariableCalledB;
         veryVeryVeryVeryVeryLongFunctionCalledA(
             veryVeryVeryVeryVeryLongVariableCalledB
-        ) +
-            c;
+        ) + c;
         veryVeryVeryVeryVeryLongFunctionCalledA(
             veryVeryVeryVeryVeryLongVariableCalledB
         ) +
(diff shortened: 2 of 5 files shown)


Janther:
prettier js formats the following:
```Javascript
veryVeryVeryVeryVeryLongVariableCalledA - veryVeryVeryVeryVeryLongVariableCalledB - 1 -1 -1 - 1 - 1 - 1 - 1;
```
like,
```Javascript
veryVeryVeryVeryVeryLongVariableCalledA -
  veryVeryVeryVeryVeryLongVariableCalledB -
  1 -
  1 -
  1 -
  1 -
  1 -
  1 -
  1;
```
have a look at #175 to see other ways prettier js groups ecuations that we are reviewing right now for solidity.

jablko:
My mistake, however it formats
```JavaScript
a(veryVeryVeryVeryVeryLongVariableCalledA + veryVeryVeryVeryVeryLongVariableCalledB) - 1;
```
like
```JavaScript
a(
  veryVeryVeryVeryVeryLongVariableCalledA +
) - 1;
```

Would you consider something like the revised version of this pull request, which handles both
```Solidity
veryVeryVeryVeryVeryLongVariableCalledA -
  veryVeryVeryVeryVeryLongVariableCalledB -
  1 -
  1 -
  1 -
  1 -
  1 -
  1 -
  1;
```
and
```Solidity
a(
  veryVeryVeryVeryVeryLongVariableCalledA +
) - 1;
```
?

Janther:
hahaha this one is a weird one.
If I try this in JS:
```Javascript
2 - a(veryVeryVeryVeryVeryLongVariableCalledA + veryVeryVeryVeryVeryLongVariableCalledB) - 1;

a(veryVeryVeryVeryVeryLongVariableCalledA + veryVeryVeryVeryVeryLongVariableCalledB) - 1 + 2;
```
I get
```Javascript
2 -
  a(
    veryVeryVeryVeryVeryLongVariableCalledA +
  ) -
  1;

a(
  veryVeryVeryVeryVeryLongVariableCalledA +
) -
  1 +
  2;
```

I like having this particular edge case as long as the main rule is not broken, could you add test cases??
testing not only your solution but also combining it in different scenarios to check exactly how it behaves not just in a specific case but also in something broader.

src/binary-operator-printers/comparison.js
```diff
@@ -23,14 +23,16 @@ module.exports = {
     const groupIfNecessary = groupIfNecessaryBuilder(path);
     const indentIfNecessary = indentIfNecessaryBuilder(path);
 
+    const right = concat([node.operator, line, path.call(print, 'right')]);
+    const parent = path.getParentNode();
+    const shouldGroup =
+      parent.type !== node.type && node.left.type !== node.type;
```
Janther: Could you also document why this particular type mismatch deserves a grouping?

jablko: :+1: Will do!

jablko:
I've rebased on #182: I added tests for
- single operator/two operands
- multiple operators
- arithmetic, comparison, and logical operators
- no context, assignment, if statement, function call, and return statement

mattiaerre:
how can we fix this cov report issue? I'll try to reach out to the Codecov team and see what they say

mattiaerre:
can you resolve the conflicts here too @jablko ? thanks

jablko:
@mattiaerre Absolutely: Rebased.

mattiaerre:
thanks os much @jablko

tests/BinaryOperators/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -330,8 +326,7 @@ contract ArithmeticOperators {
                 veryVeryVeryVeryVeryLongVariableCalledB;
         veryVeryVeryVeryVeryLongFunctionCalledA(
             veryVeryVeryVeryVeryLongVariableCalledB
-        ) **
-            c;
+        )**c;
```
mattiaerre: just noticing that there is no space before `**` and after it; is that what we want?

Janther: this was applied in #174

mattiaerre: can we approve this PR?

PR #182, opened 2019-08-05
Don't indent the leftmost operand

jablko:
Like [language-js](https://github.com/prettier/prettier/blob/master/src/language-js/printer-estree.js#L645).

### DIFF
--- a/src/binary-operator-printers/arithmetic.js
+++ b/src/binary-operator-printers/arithmetic.js
@@ -17,14 +17,18 @@ const groupIfNecessaryBuilder = path => doc => {
 };
 
 const indentIfNecessaryBuilder = path => doc => {
-  const parentNode = path.getParentNode();
-  if (
-    parentNode.type === 'BinaryOperation' &&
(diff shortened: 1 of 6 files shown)


Janther:
Hi @jablko, could you add the scenario in the tests please?

Janther:
also as far as I understand the leftmost operand is already not being indented so if I could see an example of this case not being applied.

jablko:
> Hi @jablko, could you add the scenario in the tests please?

Definitely: I've now added tests. It avoids formatting
```Solidity
veryVeryVeryVeryVeryLongFunctionCalledA(
 ) +
    c;
```
as
```Solidity
veryVeryVeryVeryVeryLongFunctionCalledA(
    ) +
    c;
```

Janther:
@jablko, could you please include this change in `src/binary-operator-printers/exponentiation.js` which was created in #174

jablko:
> could you please include this change in `src/binary-operator-printers/exponentiation.js` which was created in #174

Sure thing: Done.

PR #190, opened 2019-08-07
Apply singleQuote option to hex literals

jablko:
`hex'DeadBeef'` -> `hex"DeadBeef"`

### DIFF
--- a/src/nodes/HexLiteral.js
+++ b/src/nodes/HexLiteral.js
@@ -1,5 +1,16 @@
+const {
+  doc: {
+    builders: { concat }
+  }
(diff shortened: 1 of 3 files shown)


mattiaerre:
@jablko do you mind to add a description when submitting a PR? also the red square in the cov report is because you have created a PR out of a fork (@fvictorio @Janther I think that's the case eh?)

Janther:
Yes, that red corner is due to that weird bug of codecov

@jablko if you are uninspired to write a lengthy description, a snippet of
before and after the PR should be enough.

jablko:
:+1: Can do!

PR #191, opened 2019-08-07
Line break before declaration initial value

jablko:
### Before

```Solidity
string veryVeryVeryVeryVeryLongVariableCalledA = "The quick brown fox jumps over the lazy dog";
```

### After

```Solidity
string veryVeryVeryVeryVeryLongVariableCalledA =
    "The quick brown fox jumps over the lazy dog";
```

### DIFF
--- a/tests/Arrays/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/Arrays/__snapshots__/jsfmt.spec.js.snap
@@ -10,20 +10,21 @@ contract Arrays {
 pragma solidity ^0.5.2;
 
 contract Arrays {
-    bytes32[] public permissions = [
-        bytes32("addPartner"),
-        bytes32("removePartner"),
-        bytes32("addGroup"),
-        bytes32("removeGroup"),
-        bytes32("addUserToGroup"),
-        bytes32("removeUserFromGroup"),
-        bytes32("addSolution"),
-        bytes32("removeSolution"),
-        bytes32("addLibrary"),
-        bytes32("removeLibrary"),
-        bytes32("addPermissionToGroup"),
-        bytes32("removePermissionFromGroup")
-    ];
+    bytes32[] public permissions =
+        [
+            bytes32("addPartner"),
+            bytes32("removePartner"),
+            bytes32("addGroup"),
+            bytes32("removeGroup"),
+            bytes32("addUserToGroup"),
+            bytes32("removeUserFromGroup"),
+            bytes32("addSolution"),
+            bytes32("removeSolution"),
+            bytes32("addLibrary"),
+            bytes32("removeLibrary"),
+            bytes32("addPermissionToGroup"),
+            bytes32("removePermissionFromGroup")
+        ];
 }
 
 `;
--- a/tests/Conditional/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/Conditional/__snapshots__/jsfmt.spec.js.snap
@@ -23,9 +23,8 @@ pragma solidity ^0.4.24;
 
 contract Conditional {
     function foo() {
-        address contextAddress = longAddress_ == address(0)
-            ? msg.sender
-            : currentContextAddress_;
+        address contextAddress =
+            longAddress_ == address(0) ? msg.sender : currentContextAddress_;
     }
 
     // TODO: work with a break in the condition level.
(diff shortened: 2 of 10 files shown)


mattiaerre:
same comment as https://github.com/prettier-solidity/prettier-plugin-solidity/pull/190#issuecomment-519379159

jablko:
### Before

```Solidity
string veryVeryVeryVeryVeryLongVariableCalledA = "The quick brown fox jumps over the lazy dog";
```

### After

```Solidity
string veryVeryVeryVeryVeryLongVariableCalledA =
    "The quick brown fox jumps over the lazy dog";
```

tests/Arrays/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -10,20 +10,21 @@ contract Arrays {
...
-        bytes32("removeUserFromGroup"),
-        bytes32("addSolution"),
-        bytes32("removeSolution"),
-        bytes32("addLibrary"),
-        bytes32("removeLibrary"),
-        bytes32("addPermissionToGroup"),
-        bytes32("removePermissionFromGroup")
-    ];
+    bytes32[] public permissions =
+        [
```
fvictorio: Are we sure we want this? I kind of prefer the previous version.

tests/SplittableCommodity/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -80,8 +80,8 @@ contract SplittableCommodity is MintableCommodity {
         public
         whenNotPaused
     {
-        address supplierProxy = IContractRegistry(contractRegistry)
-            .getLatestProxyAddr("Supplier");
+        address supplierProxy =
+            IContractRegistry(contractRegistry).getLatestProxyAddr("Supplier");
```
fvictorio: Same comment as before.

mattiaerre:
can you resolve the conflicts here too @jablko ? thanks

tests/Arrays/__snapshots__/jsfmt.spec.js.snap
mattiaerre: I agree w/ that

tests/Conditional/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -23,9 +23,8 @@ pragma solidity ^0.4.24;
 
 contract Conditional {
     function foo() {
-        address contextAddress = longAddress_ == address(0)
-            ? msg.sender
-            : currentContextAddress_;
+        address contextAddress =
+            longAddress_ == address(0) ? msg.sender : currentContextAddress_;
```
mattiaerre: I like the previous version more

tests/FunctionCalls/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -23,8 +23,8 @@ contract FunctionCalls {
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 contract FunctionCalls {
     function foo() {
-        address veryLongValidatorAddress = veryVeryVeryLongSignature
-            .popLast20Bytes();
+        address veryLongValidatorAddress =
+            veryVeryVeryLongSignature.popLast20Bytes();
```
mattiaerre: this one is nicer for me in this case

mattiaerre:
what do we wanna do w/ this PR @Janther @fvictorio @jablko ?

fvictorio:
With these many conflicts, maybe it's easier to just do a new PR for this.

Janther:
Personally I don't think that breaking splitting after the `=` sign is something we want.

fvictorio:
> Personally I don't think that breaking splitting after the = sign is something we want.

FWIW prettier-js does it, if I remember correctly. Like with many things prettier does, I hated it at the beginning, but eventually got used to it.

Janther:
> FWIW prettier-js does it, if I remember correctly. Like with many things prettier does, I hated it at the beginning, but eventually got used to it.

The way prettier does it is very specific. It only indents the values if the expression in the right is a binary operation. This one will break if it's a long function call as well.

mattiaerre:
I'd be inclined to close this PR; what do you think @fvictorio @Janther ?

fvictorio:
Yes, it's been a while and it has several merge conflicts. At this point, assuming this is something we want, it's just easier to make a new PR.

PR #193, opened 2019-08-13
Moving the quote logic from the printer to the parser.

Janther:
the AST is slightly different when printing with a different quote.
So instead of deciding with which quote to print, we decide with which quote to parse.


mattiaerre:
I ❤️ this idea, everybody's happy!

Issue #207, opened 2019-11-04
Brackets in function invocation error out


Alonski:
Error:
```
contracts/examples/RockPaperScissors.sol[error] contracts/examples/RockPaperScissors.sol: ParserError: missing ';' at '(' (158:61)
[error]     at Object.parse (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/solidity-parser-antlr/dist/index.js:79:11)
[error]     at Object.parse (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier-plugin-solidity/src/parser.js:16:25)
[error]     at Object.parse$2 [as parse] (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:10631:19)
[error]     at coreFormat (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:13890:23)
[error]     at format (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:14148:73)
[error]     at formatWithCursor (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:14164:12)
[error]     at Object.formatWithCursor (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:42758:15)
[error]     at format$1 (/mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:44127:21)
[error]     at /mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:44322:16
[error]     at /mnt/d/Projects/fun-stuff/state-channels/monorepo/node_modules/prettier/bin-prettier.js:44262:14
```
This line doesn't pass Prettier and errors:
```solidity
Outcome.OutcomeItem[] memory outcomeFrom = abi.decode(fromPart.outcome, (Outcome.OutcomeItem[]));
```
This works but is incorrect:
```solidity
Outcome.OutcomeItem[] memory outcomeFrom = abi.decode(fromPart.outcome, (Outcome.OutcomeItem));
```

Relevant file: https://github.com/statechannels/monorepo/blob/alon/feature/rps-app/packages/nitro-protocol/contracts/examples/RockPaperScissors.sol#L158

[...]

PR #208, opened 2019-11-09
feat: Create failing test for array[] used in function call

Alonski:
Demonstrates #207


mattiaerre:
thanks @Alonski hopefully `solidity-parser-antlr@0.5.0` will be released soon

// cc @federicobond @fvictorio

Janther:
the test here is not valid, I made a PR with the exact code proposed in #207

PR #209, opened 2019-12-23
true and false Expression need to be indented in case the line breaks

Janther:
before:
```Solidity
        asset == ETH
            ? require(
            address(uint160(msg.sender)).send(quantity),
            "failed to transfer ether"
        ) // explicit casting to \`address payable\`
            : transferTokensToAccountSecurely(
            Token(asset),
            quantity,
            msg.sender
        );
```

after:
```Solidity
        asset == ETH
            ? require(
                address(uint160(msg.sender)).send(quantity),
                "failed to transfer ether"
            ) // explicit casting to \`address payable\`
            : transferTokensToAccountSecurely(
                Token(asset),
                quantity,
                msg.sender
            );
```

### DIFF
--- a/src/nodes/Conditional.js
+++ b/src/nodes/Conditional.js
@@ -9,12 +9,8 @@ const Conditional = {
     group(
       concat([
         path.call(print, 'condition'),
-        indent(line),
-        '? ',
-        path.call(print, 'trueExpression'),
-        indent(line),
-        ': ',
-        path.call(print, 'falseExpression')
+        indent(concat([line, '? ', path.call(print, 'trueExpression')])),
+        indent(concat([line, ': ', path.call(print, 'falseExpression')]))
       ])
     )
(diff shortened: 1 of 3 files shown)


src/nodes/Conditional.js
```diff
@@ -9,12 +9,8 @@ const Conditional = {
     group(
       concat([
         path.call(print, 'condition'),
-        indent(line),
-        '? ',
-        path.call(print, 'trueExpression'),
-        indent(line),
-        ': ',
-        path.call(print, 'falseExpression')
+        indent(concat([line, '? ', path.call(print, 'trueExpression')])),
```
fvictorio: Does it make sense to do `[line, '?', line, ...]` here?

Janther: it does. Thank you for spotting this

PR #212, opened 2019-12-26
adding commas between arguments and indenting the arguments

### LINKED ISSUE #211: Prettier produce an invalid code when passing arguments for a base class using a class notation
The following code prettifed:

```solidity
contract Parent {
    constructor(uint256 foo, bool bar) public {

    }
}

contract Child is Parent(5456, false) {
  
}
```

Results to this
```solidity
contract Parent {
    constructor(uint256 foo, bool bar) public {}
}

contract Child is Parent(5456false) {}

```
which is a faulty code.

Janther:
Fixes #211

### DIFF
--- a/src/nodes/InheritanceSpecifier.js
+++ b/src/nodes/InheritanceSpecifier.js
@@ -1,21 +1,32 @@
 const {
   doc: {
-    builders: { concat }
+    builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier/standalone');
 
-const InheritanceSpecifier = {
-  print: ({ node, path, print }) => {
-    let parts = [path.call(print, 'baseName')];
-
-    if (node.arguments && node.arguments.length) {
-      parts.push('(');
-      parts = parts.concat(path.map(print, 'arguments'));
-      parts.push(')');
-    }
-
-    return concat(parts);
+const printArguments = (node, path, print) => {
+  if (node.arguments && node.arguments.length) {
+    return group(
+      concat([
+        '(',
+        indent(
+          concat([
+            softline,
+            join(concat([',', line]), path.map(print, 'arguments'))
+          ])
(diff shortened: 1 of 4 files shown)


src/nodes/InheritanceSpecifier.js
```diff
@@ -1,21 +1,32 @@
...
-    builders: { concat }
+    builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier/standalone');
 
-const InheritanceSpecifier = {
-  print: ({ node, path, print }) => {
-    let parts = [path.call(print, 'baseName')];
+const printArguments = (node, path, print) => {
+  if (node.arguments && node.arguments.length) {
```
fvictorio: I wonder what happens (and what should happen) if someone does `Foo is Bar()`. I think this will format it as `Foo is Bar`. I'm pretty sure that's OK, but just something to keep in mind.

Janther: A quick check using https://astexplorer.net tells me that we will be removing the `()`.
This will generate a difference in the AST.
I believe that in this case we should be opinionated and remove the `()` anyway.
If we agree on this I'll add this rule so we can add this example and still pass the `AST_COMPARE=true` tests.

Janther: Forget what I said.
`Foo is Bar` and `Foo is Bar()` create the same AST.
Thus we can't know from the AST alone.
I still believe that we should just ignore the `()`,  we could check the original source but that will be extra work that it's not really needed.

fvictorio: +1 to removing it, especially if the AST is the same.

mattiaerre: I'll merge and deploy this one if that's ok

PR #216, opened 2020-01-24
Removing extra empty lines at the end of a block

Janther:
before we preserved 1 empty line  if the developer originally had it with the exception of the first line. This would lead to this input:
```Solidity
contract A {

    function a() public {}

    function b() public {}

    function c() public {}

}
```

to be formatted as:
```Solidity
contract A {
    function a() public {}

    function b() public {}

    function c() public {}

}
```

The last line always bugged me so I decided to remove it, now it will look like this:

```Solidity
contract A {
    function a() public {}

    function b() public {}

    function c() public {}
}
```


Janther:
closing since it's part of #217

PR #217, opened 2020-01-24
Adding spaces between functions and contracts

Janther:
Following the style guide:
  - contracts, libraries, and interfaces should be separated by 2 empty lines.
  - functions should be separated by 1 empty line.

This PR solves this in most cases automatically.


fvictorio:
Just for the record, this was something that was _removed_ in the past: https://github.com/prettier-solidity/prettier-plugin-solidity/pull/103 Not saying that that's the correct option, but at the very least this is a place where we need to choose between the style guide and prettier's philosophy. I personally am on the fence on this.

Janther:
Fair enough, we can discuss it here before we merge.
I like that this is enforcing the style guide though and being a little more compatible with `solhint`.
This PR solves 3 things that have been bugging me.
  - the empty line at the end of a block
  - functions are surrounded by 1 empty line
  - contracts are surrounded by 2 empty lines (Can be changed to 1)

We are still preserving empty spaces along the code (with the exception of the end and beginning of blocks)

mattiaerre:
happy to merge this personally, shall I proceed @Janther @federicobond ?

Janther:
I’m happy to merge this.
It’s easy to change 2 lines for 1 if the community doesn’t like it. But I
feel they’ll understand that this is in solidity’s best practices.

fvictorio:
Then it's OK with me :+1:

fvictorio:
> I like that this is enforcing the style guide though and being a little more compatible with `solhint`.

Just for the record, we are going to drop all of solhint's style rules in 3.0. My goal is for this plugin to be responsible of the styling (through [`solhint-plugin-prettier`](https://github.com/fvictorio/solhint-plugin-prettier)).

PR #220, opened 2020-01-31
Add support for solidity 0.6.x

### LINKED ISSUE #218: Support for Solidity 0.6
Consensys forked and plans to maintain its own version of `solidity-parser-antlr` (discussion: https://github.com/federicobond/solidity-parser-antlr/issues/98). I'm not sure what Federico's plans are (not tagging his username to avoid spamming him).

At Solhint we are using a *temporary* fork I made, with both Consensys updates and some changes we needed: https://www.npmjs.com/package/@fvictorio/solidity-parser

Anyway, at some point we'll need to upgrade the parser. I opened this issue as a reminder and to explain the situation.

fvictorio:
Closes #218 

- Change parser for consensys's fork
- Add basic support for try/catch

There are other new features I'm not sure are supported. See the list [here](https://solidity.readthedocs.io/en/v0.6.0/060-breaking-changes.html#new-features). For example, I haven't checked if there's something extra we need to do for the `leave` assembly statement.


Janther:
based on the tests that they have, we are still not implementing these nodes:
- `AssemblyFunctionDefinition`
- `AssemblyStackAssignment`
- `DoWhileStatement`

Also, it seems they are not testing for these nodes:
- `BreakStatement`
- `Conditional`
- `ContinueStatement`
- `HexLiteral`
- `IndexAccess`

From this list, only the `ContinueStatement` is not on our coverage so we need to add it and check that is still used by the parser.

So here's a list of things still needed for this PR.
- [ ] Add node for `AssemblyFunctionDefinition`
- [ ] Add node for `AssemblyStackAssignment `
- [ ] Add node for `DoWhileStatement `
- [ ] Add test for `ContinueStatement `

Sorry, I don't have time right now to help but I think the research was already a big one ;)

src/nodes/index.js
```diff
@@ -16,12 +16,12 @@ module.exports = {
   Block: require('./Block.js'),
   BooleanLiteral: require('./BooleanLiteral.js'),
   BreakStatement: require('./BreakStatement.js'),
+  CatchClause: require('./CatchClause.js'),
```
mattiaerre: this should be automatically generated by running the `generateindexes` script

package.json
```diff
@@ -80,6 +80,7 @@
     "prettier": "^1.15.3",
     "semver": "^6.3.0",
     "solidity-parser-antlr": "^0.4.11",
```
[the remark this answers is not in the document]
fvictorio: Done.

src/nodes/index.js
fvictorio: I ran this but the file didn't change except for the timestamp, so I guess it doesn't make sense to push that.

mattiaerre: no I mean you added the new parts by hand while instead you should have just run the script or am I missing something?

fvictorio:
> Sorry, I don't have time right now to help but I think the research was already a big one ;)

That wasn't big, it was huge, thank you!

I did a couple of them, will work on the rest as soon as I can. My `AssemblyFunctionDefinition` is not perfect (I'm not breaking the return parameters) but I think it's a good start.

src/nodes/index.js
mattiaerre: thanks, now we know this has been generated automatically 👍

fvictorio:
I'll postpone the `AssemblyStackAssignment` part until while this issue is open: https://github.com/ConsenSys/solidity-parser-antlr/issues/7

fvictorio:
Added support for do-while statements. Not really ideal yet. For example, this:

```
do  a = a.add(LONG_VARIABLE).add(LONG_VARIABLE).add(LONG_VARIABLE); while (a < 200);
```

is formatted like this

```
        do
            a = a.add(LONG_VARIABLE).add(LONG_VARIABLE).add(
                LONG_VARIABLE
            ); while (a < 200);
```

I don't know what the ideal output should be, but at the very least I'd say that if a do-while has a single statement, and if a single statement will be broken, then braces should be added. Of course, if we do that here, we should do it in the WhileStatement too.

fvictorio:
That being said, I would just merge and publish this, and open an issue for the stack assignment.

fvictorio:
#227

PR #229, opened 2020-02-18
ElementaryTypeName takes care of how to print state mutability

### LINKED ISSUE #226: Removed wrong keywords when auto-formatting
The address payable keyword is wrongly corrected as address when using the helper contract Address.sol from OpenZeppelin. What I had in the smart contract was like

```using Address for address payable;```
But got wrongly auto-corrected as

```using Address for address;```

This issue was raised github.com/juanfranblanco/vscode-solidity/issues/162 by @maxweng

### PR DESCRIPTION
**Pull request by Janther**
fixes #226

Before the PR we were manually choosing where to print state mutability in `ArrayTypeName` and `VariableDeclaration`.
Now we rely on `ElementaryTypeName` to do that for us.
plus a small refactor on `VariableDeclaration`

### DIFF
--- a/src/nodes/VariableDeclaration.js
+++ b/src/nodes/VariableDeclaration.js
@@ -1,18 +1,24 @@
 const {
   doc: {
-    builders: { join }
+    builders: { concat }
   }
 } = require('prettier/standalone');
 
-const indexed = node => (node.isIndexed ? 'indexed' : '');
+const indexed = node => (node.isIndexed ? ' indexed' : '');
 
 const visibility = node =>
-  node.visibility !== 'default' ? node.visibility : '';
+  node.visibility && node.visibility !== 'default'
+    ? concat([' ', node.visibility])
+    : '';
 
-const constantKeyword = node => (node.isDeclaredConst ? 'constant' : '');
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: src/nodes/VariableDeclaration.js [hunk 1]
```diff
@@ -1,38 +1,39 @@
...
+
+const visibility = node =>
+  node.visibility && node.visibility !== 'default'
+    ? concat([' ', node.visibility])
+    : '';
+
+const constantKeyword = node => (node.isDeclaredConst ? ' constant' : '');
+
+const storageLocation = node =>
+  node.storageLocation && node.visibility !== 'default'
```
Janther: This is the only place where I think the code is a bit weird by asking whether `node.visibility !== 'default'` to make a choice for `storageLocation` but I took it from the original code.
What do you think @fvictorio @mattiaerre ??

fvictorio: I think this makes sense, because `node.visibility === 'default'` means there's no explicit visibility in the original code.

mattiaerre: I agree @Janther why is visibility necessary to determine the location? it looks odd to me too. But it seems to work so I'll go ahead, approve and merge

PR #230, opened 2020-02-18
Splitting and indenting variable declarations when necesary

### LINKED ISSUE #225: long tuple assignment formatted to single line
say, I have a long tuple declaration and assignment in solidity code.
It's convenient to have it formatted like long list of arguments for method - one method per line, i.e.:

`(string memory holderId,`  
`string memory containerId,`  
`string memory parentContainerId,`  
`bytes32 volumeId,`  
`bytes32 subUnitId,`  
`bytes32 requestId) = getParamsById(paramsId);`

But prettier-plugin converts tuple to single line, like:

`(string memory holderId,string memory containerId, string memory parentContainerId, bytes32 volumeId, bytes32 subUnitId, bytes32 requestId) = getParamsById(`  
`    paramsId`  
`);`

It even violates previously set line width, which makes it very uncomfortable to read.
Is there any setting for this or this is a bug/feature?

Janther:
Fixes #225

Before PR:
```Solidity
(string memory holderId,string memory containerId, string memory parentContainerId, bytes32 volumeId, bytes32 subUnitId, bytes32 requestId) = getParamsById(
    paramsId
);
```

After PR:
```Solidity
(
    string memory holderId,
    string memory containerId,
    string memory parentContainerId,
    bytes32 volumeId,
    bytes32 subUnitId,
    bytes32 requestId
) = getParamsById(paramsId);
```

### DIFF
--- a/src/nodes/VariableDeclarationStatement.js
+++ b/src/nodes/VariableDeclarationStatement.js
@@ -1,29 +1,43 @@
 const {
   doc: {
-    builders: { concat, join }
+    builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier/standalone');
+
+const embraceVariables = (doc, embrace) =>
+  embrace ? concat(['(', doc, ')']) : doc;
+
+const variables = (node, path, print) =>
+  group(
+    concat([
+      indent(
+        concat([
+          softline,
+          join(
+            concat([',', line]),
+            path.map(statementPath => print(statementPath), 'variables')
+          )
+        ])
+      ),
+      softline
+    ])
+  );
+
+const initialValue = (node, path, print) =>
+  node.initialValue ? concat([' = ', path.call(print, 'initialValue')]) : '';
 
 const VariableDeclarationStatement = {
   print: ({ node, path, print }) => {
     const startsWithVar =
       node.variables.filter(x => x && x.typeName).length === 0;
 
-    let doc = join(
-      ', ',
-      path.map(statementPath => print(statementPath), 'variables')
-    );
-
-    if (node.variables.length > 1 || startsWithVar) {
-      doc = concat(['(', doc, ')']);
-    }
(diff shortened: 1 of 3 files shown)


src/nodes/VariableDeclarationStatement.js
```diff
@@ -1,29 +1,43 @@
...
-    builders: { concat, join }
+    builders: { concat, group, indent, join, line, softline }
   }
 } = require('prettier/standalone');
 
+const embraceVariables = (doc, embrace) =>
+  embrace ? concat(['(', doc, ')']) : doc;
+
+const variables = (node, path, print) =>
+  group(
```
fvictorio: I wonder if we can (and if we should) make a helper for this pattern, since we use it a lot. I guess something like this:

```
const printList = (node, path, print, key, separator = ',') => 
  group(
    concat([
      indent(
        concat([
          softline,
          join(
            concat([separator, line]),
            path.map(statementPath => print(statementPath), key)
          )
        ])
      ),
      softline
    ])
  );
```

(with better names for the function and the parameters).

Janther: This is a tricky one.
The pattern is used a lot but with small differences.
Sometimes times is a `softline` and sometimes is a `line`, the separator and the content also changes I’ll see how it looks having a function with many variables in one go vs a step by step declaration.
 The system is already modular and there aren’t many files bigger than a screen.
I’ll give it a thought and play with the idea a little but probably for another PR.

fvictorio: Sure. Just for the record: I'm fine with merging this as it is, I was just thinking out loud.

Janther: Me too, I already got over my phase of having the driest possible code without really thinking of maintainability

PR #234, opened 2020-02-19
Print list

Janther:
just playing with the idea of having a `printList` function (provided by @fvictorio) for most of our splitting and indenting cases.

The function just needs an array, but can also receive a `firstSeparator` (usually a `softline` used to separate from the previous character), a normal `separator` to put between items (usually a `concat([',', line])`), and finally the `lastSeparator` that in most cases is the same as `firstSeparator` with 2 exceptions (`AssemblySwitch` and `StructDefinition`).

```Javascript
const printList = (
  list,
  firstSeparator = softline,
  separator = concat([',', line]),
  lastSeparator = firstSeparator
) =>
  group(
    concat([
      indent(concat([firstSeparator, join(separator, list)])),
      lastSeparator
    ])
  );
```

Please have a look and see if it makes sense to use this pattern.

### DIFF
--- a/src/nodes/AssemblyCall.js
+++ b/src/nodes/AssemblyCall.js
@@ -1,31 +1,21 @@
 const {
   doc: {
-    builders: { concat, group, indent, join, line, softline }
+    builders: { concat }
   }
 } = require('prettier/standalone');
 
+const printList = require('./print-list');
+
 const AssemblyCall = {
-  print: ({ node, path, print }) => {
-    if (node.arguments.length === 0) {
-      return node.functionName;
-    }
-    return concat([
-      node.functionName,
-      '(',
-      group(
-        concat([
-          indent(
-            concat([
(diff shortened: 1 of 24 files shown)


fvictorio:
I think I like this. Most usages use default arguments, so that's nice.

The only problem I see is that when explicit arguments are added, it gets a little cryptic. What do you think about having just an `options` parameter, so that the calls are more explicit (the "javascript doesn't have named parameters so this is the best we can do" pattern):

```
const printList = (
  list,
  {
    firstSeparator = softline,
    separator = concat([',', line]),
    lastSeparator = firstSeparator,
  } = {}
) =>
```

Janther:
done, this idea was on the back of my mind. Glad that you proposed it aswell.

mattiaerre:
this is really cool, let's merge and deploy a new version

PR #236, opened 2020-02-24
1.0.0-alpha.41

mattiaerre:
# Description

- bump version


Janther:
could we do #237 before bumping the version??

mattiaerre:
absolutely @Janther I'll wait 👍

PR #237, opened 2020-02-24
Coverage

Janther:
This PR solves an issue with try/catch discovered by adding more coverage cases and ups the coverage of `src/nodes` to a 100%

### DIFF
--- a/src/nodes/CatchClause.js
+++ b/src/nodes/CatchClause.js
@@ -6,15 +6,22 @@ const {
 
 const printList = require('./print-list');
 
+const parameters = (node, path, print) =>
+  node.parameters
+    ? concat([
+        node.isReasonStringType ? 'Error' : '',
+        '(',
+        printList(path.map(print, 'parameters')),
+        ') '
+      ])
+    : '';
+
 const CatchClause = {
   print: ({ node, path, print }) => {
     return group(
       concat([
         'catch ',
-        node.isReasonStringType ? 'Error' : '',
-        '(',
-        printList(path.map(print, 'parameters')),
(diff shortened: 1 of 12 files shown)


src/parser.js
```diff
@@ -21,9 +21,7 @@ function parse(text, parsers, options) {
       if (ctx.initExpression) {
         ctx.initExpression.omitSemicolon = true;
       }
-      if (ctx.loopExpression) {
-        ctx.loopExpression.omitSemicolon = true;
-      }
+      ctx.loopExpression.omitSemicolon = true;
```
mattiaerre: I guess that w/in `ForStatement`s we always have a `ctx.loopExpression` right?

src/nodes/CatchClause.js
```diff
@@ -6,15 +6,22 @@ const {
 
 const printList = require('./print-list');
 
+const parameters = (node, path, print) =>
```
mattiaerre: just noticing this right now, do we always use function expressions?

src/parser.js
Janther: Yeah, the initial expression and the condition expression can be `null` but for some reason the loop expression is an expression with `null` content.
Maybe we can ask for some consistency but right now that’s how it is

src/nodes/CatchClause.js
Janther: I tend to favour this notation. Is there an improvement by using declarations instead?

mattiaerre: not really, but to me, if it's a function I like to signal that w/ the `function` keyword

Issue #238, opened 2020-02-25
Inline assembly change in solidity 0.6: handle new representation for built-in opcodes such as `gas`


nicholaspai:
[Solidity 0.6 Handles Inline Assembly opcodes differently](https://solidity.readthedocs.io/en/v0.6.0/assembly.html)

For example, `gas` is now `gas()`, which might be used in the smart contract method:

```
function _executeCall(address to, uint256 value, bytes memory data) private returns (bool success) {
        // Mostly copied from:
        // solhint-disable-next-line max-line-length
        // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
        // solhint-disable-next-line no-inline-assembly

        // Ignoring this line for now because solidity 0.6 changed the "gas" instruction to "gas()", which prettier does not accept
        // prettier-ignore
        assembly {
            let inputData := add(data, 0x20)
            let inputDataSize := mload(data)
            success := call(gas(), to, value, inputData, inputDataSize, 0, 0)
        }
    }
```

Note that my current workaround is to just `prettier ignore` the entire AssemblyBlock

fvictorio:
Does this mean that we should do different things if the `pragma` directive is >= 0.6? I don't think we have support for that, but it should be doable. This could also be an option in the prettier configuration, but that seems a little weird to me.

Thoughts @mattiaerre, @Janther?

Janther:
Made a small PR handling this case.
I'm a bit unfamiliar with how developers behave in assembly but feel it's fine like this.
maybe someone could check the compiled code for both cases?

nicholaspai:
> I'm a bit unfamiliar with how developers behave in assembly but feel it's fine like this.
> maybe someone could check the compiled code for both cases?

Nice work. Not entirely sure what you’re asking here, but as far as I can tell, they made opcodes that didn’t take arguments into zero-input “functions”

https://solidity.readthedocs.io/en/v0.6.0/assembly.html
https://solidity.readthedocs.io/en/v0.5.3/assembly.html

fvictorio:
As far as I understand this, before 0.6 you could use either `gas` or `gas()`. In 0.6 you can only use `gas()`. Is this correct?

[...]

PR #239, opened 2020-02-27
Assembly function calls without parameters

Janther:
Solves #238 
Check if the last character is a ')' to catch function calls without parameters

### DIFF
--- a/src/nodes/AssemblyCall.js
+++ b/src/nodes/AssemblyCall.js
@@ -7,8 +7,9 @@ const {
 const printList = require('./print-list');
 
 const AssemblyCall = {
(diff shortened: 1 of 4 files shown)


fvictorio:
I wonder if this is something that the parser should handle? I just implemented the override functionality and what I did was:

- no override: `override: null`
- standalone override: `override: []`
- override with base classes: `override: [{...}, {...}]`

In this case, maybe a better API would be to have null arguments when there are no parentheses. Of course, this is a breaking change for the parser, but maybe it's worth opening an issue. (FWIW: this comment doesn't block this PR at all)

PR #240, opened 2020-02-28
favouring terciary operator over if...return...else...return

Janther:
Not really necessary. It just feels nicer and a bit cleaner.


mattiaerre:
Oh, that's a good point @fvictorio I don't wanna compromise readability. But in this case, I'm neutral. What should we be doing?

fvictorio:
I'm mostly neutral too, so I'm fine with merging it.

PR #241, opened 2020-02-28
1.0.0-alpha.42

mattiaerre:
# Description

- bump version

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,7 @@
   "env": {
     "jest": true
   },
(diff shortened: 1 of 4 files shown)


Janther:
node 8 reached the end of it's life last year.
eslint dropped support [2 months ago](https://github.com/eslint/eslint/commit/2c28fbbb563a44282bef0c9fcc9be29d611cc83b)
we should too.

mattiaerre:
I agree @Janther however I'm not sure I want to support 13; let's only use 10 and 12

Issue #242, opened 2020-02-28
Handle fallback and receive functions


fvictorio:
One of the breaking changes of solidity 0.6 is that the fallback (anonymous) function now is replaced by two explicit functions: `fallback()` and `receive()`. So right now prettier is doing this for fallback:

```diff
- fallback () external {}
+ function() external {}
```

and this for receive:

```diff
- receive () external payable {}
+ function() external payable {}
```

For the record, anonymous functions are a compile error in solidity 0.6. This means that we should just keep whatever we get: if it's anonymous is kept anonymous, ditto for explicit fallback and receive. The question is if the parser will let us tell apart between the different possibilities.

PR #244, opened 2020-02-28
receive and fallback support

Janther:
solves #242
`receive` was easy but for `fallback` I rely on the `originalText`

### DIFF
--- a/src/nodes/FunctionDefinition.js
+++ b/src/nodes/FunctionDefinition.js
@@ -12,15 +12,18 @@ const functionName = (node, options) => {
   if (node.isReceiveEther) return 'receive';
   // The parser doesn't give us any information about the keyword used for the
   // fallback.
-  // Luckily, `function` and `fallback` have 8 characters so using the
-  // originalText is the next best option.
+  // Using the originalText is the next best option.
   //
   // An neat idea would be to rely on the pragma and enforce it but for the
   // moment this will do.
-  return options.originalText.slice(
-    options.locStart(node),
-    options.locStart(node) + 8
-  );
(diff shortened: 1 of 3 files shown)


src/nodes/FunctionDefinition.js
```diff
@@ -6,10 +6,21 @@ const {
...
-  return 'function';
+  if (node.isReceiveEther) return 'receive';
+  // The parser doesn't give us any information about the keyword used for the
+  // fallback.
+  // Luckily, `function` and `fallback` have 8 characters so using the
+  // originalText is the next best option.
+  //
+  // An neat idea would be to rely on the pragma and enforce it but for the
+  // moment this will do.
+  return options.originalText.slice(
```
mattiaerre: should we enforce here that we should return only `function` and `fallback` and not, for instance, `facepalm` (yes it's 8 letters)?

Janther: Can do that but will mess with coverage though as there’s no chance for us to trigger the parser to have `facepalm` as a function definition

Janther: ok I updated it to be a bit more explicit

mattiaerre: hahaha, I've pushed a commit, what do you think? something in between?

Janther: yup, this way we'll also know if solidity inserts a new keyword as `undefined` doesn't silently print in `concat`

mattiaerre: I go ahead and merge then ok?

Issue #187, opened 2019-08-06
Format of `for`s with missing parts


fvictorio:
An "empty" for is printed as `for (; ; )`. I think `for (;;)` would be better.

We should also consider other scenarios, like the initialization being missing (probably the most likely): `for (; i < k; i++)`.

Janther:
solved in #247

PR #247, opened 2020-03-02
Separated lists and separated items

Janther:
I've been playing with the `printList` pattern and noticed that we also use it a lot for single items for example in a condition statement.

this one is not really necesary to merge but I'd like your opinions on this and if we find it's usefull we can merge it.


Janther:
travis keeps failing to install eslint on node 10 but it's alright on the after merge build

mattiaerre:
let me connect w/ them again @Janther

Janther:
This is travis, not codecov though

mattiaerre:
yeah @Janther totally but I've had some convos w/ them in the past 👍

fvictorio:
Looks good. Maybe we should document these two functions for future collaborators and, more important, future ourselves?

mattiaerre:
that's a good idea actually @fvictorio

src/nodes/print-separated-list.js
```diff
@@ -4,6 +4,13 @@ const {
   }
 } = require('prettier/standalone');
 
+// This function will add an indentation to the `list` and separate it from the
+// rest of the `doc` in most cases by a `softline`.
+// the list itself will be printed with a separator that in most cases is a
+// comma (,) and a `line`
+//
+// NOTE: the resulting `doc` in wrapped in a `group` because multiple items
```
fvictorio:
```suggestion
// NOTE: the resulting `doc` is wrapped in a `group` because multiple items
```

Janther:
the issue with that particular test environment is that `eslint` doesn't seem to be installed. Last time I had this happened I just introduced `package-lock.json` to the repo and it was solved

mattiaerre:
legend @Janther it worked!

tests/ForStatements/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -57,9 +78,23 @@ contract ForStatements {
             a++;
         }
 
-        for (; ; ) {
+        for (;;) {
```
mattiaerre: why this one has no space in between anymore?

tests/TryCatch/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -120,9 +120,7 @@ contract FeedConsumer {
         // Permanently disable the mechanism if there are
         // more than 10 errors.
         require(errorCount < 10);
-        try
-            feed.getData(token)
-        returns (
+        try feed.getData(token) returns (
```
mattiaerre: is this ok?

tests/ForStatements/__snapshots__/jsfmt.spec.js.snap
Janther: have a look at #187

tests/TryCatch/__snapshots__/jsfmt.spec.js.snap
Janther: I like it this way better but could be convinced otherwise

tests/ForStatements/__snapshots__/jsfmt.spec.js.snap
mattiaerre: oh yes I remember, so what about the scenario in line 97 below?

tests/ForStatements/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -57,9 +78,23 @@ contract ForStatements {
...
+
+        for (i = 0; ; i++) {}
+
+        for (i = 0; i < 100; i++) {}
+
+        for (; i < 100; i++) {}
+
+        for (; i < 100; ) {}
+
+        for (; ; i++) {}
```
mattiaerre: this one I mean ^^

Janther: hehe this is based on how prettier js formats `for` statements.
When in doubt I always go to prettier js,
```Solidity
for (; ; i++) {}

for (;; i++) {}
```

The change is very easy, you can see which one you like better and we can work towards that.

Janther: I suggest you try the different combinations and see how they appeal to your eyes. I did this based on what prettierJS does but I also tried others and just didn't like it as I got lost at a quick glance.

mattiaerre: so you prefer w/ a space in between; which is also what prettier-js does correct?

Janther: I like that there's a space after the semicolon whenever there's at least one expression in the `for` statement.
Only when the `for` is empty I like it without the spaces.
This just happens to coincide with prettier-js.

mattiaerre: sweet, let's go w/ this one then!

PR #248, opened 2020-03-02
adding test described in #207

Janther:
The new parser parses correctly the example in #207 
There's nothing for us to do here, so I'll add a test to make sure it doesn't break in the future.


mattiaerre:
I can approve and merge this one then @Janther

Janther:
Yeah there’s nothing for us to do as it’s already working. Just adding the use case in case we decide to change the parser again.

PR #257, opened 2020-04-06
Splitting params if the list is longer than 2

Janther:
The idea of splitting not only when the line is too long but also when the line is complex is something we should look more into.
Here is an idea proposed by @Jaime-Iglesias in #256

### DIFF
--- a/tests/StyleGuide/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/StyleGuide/__snapshots__/jsfmt.spec.js.snap
@@ -355,34 +355,35 @@ contract FunctionDeclaration {
         doSomething();
     }
 
-    function thisFunctionNameIsReallyLong(address x, address y, address z)
-        public
-        onlyowner
-        priced
-        returns (address)
-    {
+    function thisFunctionNameIsReallyLong(
+        address x,
+        address y,
+        address z
+    ) public onlyowner priced returns (address) {
         doSomething();
     }
 
-    function thisFunctionNameIsReallyLong(address x, address y, address z)
-        public
-        onlyowner
-        priced
-        returns (address)
-    {
+    function thisFunctionNameIsReallyLong(
+        address x,
+        address y,
+        address z
+    ) public onlyowner priced returns (address) {
         doSomething();
     }
 
-    function thisFunctionNameIsReallyLong(address x, address y, address z)
-        public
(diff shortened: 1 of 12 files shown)


mattiaerre:
I like this a lot ... consistency FTW 👍

tests/StyleGuide/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -355,34 +355,35 @@ contract FunctionDeclaration {
...
-        public
-        onlyowner
-        priced
-        returns (address)
-    {
+    function thisFunctionNameIsReallyLong(
+        address x,
+        address y,
+        address z
+    ) public onlyowner priced returns (address) {
```
fvictorio, line 378: I'm randomly sampling the test cases to see how this looks. Why is this line not being split? I guess we are looking at the number of modifiers, not counting the return and visibility parts. Should we? Especially since solidity 0.6 adds other "non-custom" modifiers that can go here, like `virtual`.

Janther: the grouping is like this:

- params
- modifiers
  - return params

so prettier will try to fit params in one line first (here is where we are forcing the break)
then it'll try to fit modifiers and return params in one line, if it breaks it'll still try to fit return params in its own line. However, if return params breaks, then modifiers will also break but initial params are in a different group.

I hope this is clear enough.

This PR is only for the number of params and return params.
Maybe we would also want a rule for number of user defined modifiers??

Janther: they are in different groups because it would be weird if a function with a single param had to split just because it had many modifiers

fvictorio: > This PR is only for the number of params and return params.
> Maybe we would also want a rule for number of user defined modifiers??

This is the part I was missing, didn't notice modifiers weren't included in the PR :sweat_smile: 

> they are in different groups because it would be weird if a function with a single param had to split just because it had many modifiers

Agree

mattiaerre:
shall we merge this PR @fvictorio ? what do you think?

fvictorio:
Yes, let's merge it.

PR #271, opened 2020-05-08
Index Range Access

### LINKED ISSUE #269: IndexRangeAccess error when contract uses Solidity 0.6 calldata array slicing
`[error] contracts/Staking.sol: Error: Unknown type: "IndexRangeAccess"`

### Demo contract
```
pragma solidity ^0.6.4;

contract Demo {
    event Log(bytes value);

    function log(bytes calldata value) external {
        emit Log(bytes(value[0:10]));
    }
}
```

### prettierrc.json 
```
{
  "printWidth": 100,
  "useTabs": false,
  "bracketSpacing": true,
  "overrides": [
    {
      "files": "*.js",
      "options": {
        "semi": false,
        "trailingComma": "all",
        "tabWidth": 2,
        "singleQuote": true,
        "explicitTypes": "always"
      }
    },
    {
      "files": "*.sol",
      "options": {
        "tabWidth": 4,
        "singleQuote": false,
        "explicitTypes": "always"
      }
    }
  ]
}
```

Janther:
closes #269


tests/IndexRangeAccess/IndexRangeAccess.sol
```diff
@@ -6,4 +6,4 @@ contract Demo {
     function log(bytes calldata value) external {
         emit Log(bytes(value[0:10]));
     }
-}
\ No newline at end of file
+}
```
Janther, line 9: hehe these files are supposed to be not very well formatted anyway so I wouldn't care much about the line at the end. but would make sure the snapshot fixes this.

mattiaerre: sorry it's my bad I can't help but fix a missing a new line at EOF 😊

PR #275, opened 2020-05-14
Adding node for NameValueExpression

Janther:
Addresses #274
Took me a while to do the array of names and the array of values elegantly @fvictorio 🤣


fvictorio:
> Took me a while to do the array of names and the array of values elegantly @fvictorio

For what it's worth, I regret doing it that way :sweat:

Issue #277, opened 2020-05-17
Override modifier getting deleted ?


ianmihura:
So, this might not be the correct place to post this. I'm coming from VScode, using this [Solidity extension](https://github.com/juanfranblanco/vscode-solidity), which seems to be using your module to format the code, [in :14](https://github.com/juanfranblanco/vscode-solidity/blob/master/src/formatter/prettierFormatter.ts)

Might be the problem is coming from another place, tho I doubt it, as the VScode extension seems to manage the `override modifier missing` error correctly, and the problem arises only as the formatter is applied to the file. 

Anyway, a similar issue already exists in the appropriate repo, [issue 174](https://github.com/juanfranblanco/vscode-solidity/issues/174), I'm only passing thru to check on you guys as well.

Cheers

abcoathup:
Using https://github.com/juanfranblanco/vscode-solidity formatting a contract with the override keyword results in the override keyword being removed
## Unformatted
```solidity
pragma solidity ^0.6.0;


contract A {
    function doStuff() public virtual {}
}


contract B is A {
    function doStuff() public override {}
}
```
## Formatted
```solidity
pragma solidity ^0.6.0;


contract A {
    function doStuff() public virtual {}
}


contract B is A {
    function doStuff() public {}
}
```

Janther:
This was solved on version [v1.0.0-alpha.49](https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-alpha.49)
Perhaps @juanfranblanco can post a newer release pointing to the latest release?

juanfranblanco:
@Janther @ianmihura yes it is is already there.. i was testing the caching feature before release.

ianmihura:
Great, thank you, amazing work 🙌

Janther:
Closing since this `vs-code` has been updated

smoove:
This problem still exists for contract state variables:

Before formatting: `mapping(address => bool) public override varName;`
After formatting: `mapping(address => bool) public varName;`

mattiaerre:
the same problem here do you think @Janther ?

Janther:
Nope, this one is on us, first I’ll check if the parser identifies
`virtual` and `override` in these cases

fvictorio:
Was this fixed by #281?

fvictorio:
Oh, modifiers still pending, sorry.

PaulRBerg:
[...]

PR #281, opened 2020-05-28
adding override keyword in state variables

Janther:
adding the case described [here](https://github.com/prettier-solidity/prettier-plugin-solidity/issues/277#issuecomment-634553637)

### DIFF
--- a/src/nodes/Block.js
+++ b/src/nodes/Block.js
@@ -1,6 +1,6 @@
 const {
   doc: {
-    builders: { concat, indent, line }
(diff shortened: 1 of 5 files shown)


Janther:
the coverage is down because I also added the case for `override` in modifiers, but we are waiting for the parser to give us that info

mattiaerre:
would this close #277 ?

Janther:
Nope, we still are missing the override and virtual keywords on modifiers

PR #287, opened 2020-06-04
Modifiers now will keep virtual and override

### LINKED ISSUE #279: Support `virtual` and `override` for modifiers.
Currently it works for functions, but not for modifiers.

Example:

```solidity
modifier onlyOwner() virtual;

modifier onlyOwner() override {
    require(msg.sender == owner(), "Ownable: caller is not the owner");
    _;
}
```

Janther:
fixes #279 and #277 
bumping the dependency and also tweaking the variable declaration node.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,28 +5,28 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.8.3",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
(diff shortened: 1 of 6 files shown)


mattiaerre:
I think I can merge #286 first @Janther what do you think?

Janther:
I also updated the parser before dependabot did, so I’m not sure what type
of conflicts will be there.
It’s up to you.

mattiaerre:
it looks like it did it before you did it and also I thought you were happy about that since you suggested to use dependabot. in this case, I'll close its PR and merge yours instead.

Janther:
Well yes but in my local machine I already had the changes

fvictorio:
Thanks for doing this @Janther!

PR #290, opened 2020-06-07
add a note in vscode section

zemse:
This node is regarding a fix introduced in prettier-vscode 3.8.0 around Nov 2019 causing vscode settings to be ignored (prettier/prettier-vscode#1079). Looks like many people around the world faced a trivial looking problem (prettier/prettier-vscode#1085).

Please let me know if any changes are required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,39 @@ You'll notice now that `prettier` is formatting every time the files are saved b
 }
 ```
 
+Note: When you install the npm package `prettier` in your project and create a `.prettierrc` file (which wasn't in your project before this), your VSCode's default settings or rules in `settings.json` are ignored ([prettier/prettier-vscode#1079](https://github.com/prettier/prettier-vscode/issues/1079)).
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -116,6 +116,39 @@ You'll notice now that `prettier` is formatting every time the files are saved b
...
+        "printWidth": 80,
+        "tabWidth": 2,
+        "useTabs": false,
+        "singleQuote": true,
+        "bracketSpacing": true
+      }
+    }
+  ]
+}
+```
```
mattiaerre, line 149: you may want to add a new line here; thanks 😊

fvictorio:
I feel we could do this more concisely?

The first paragraph added is great. After that, I think we can just say something like:

> If you want a different configuration for your javascript and solidity files, you can add an [overrides property](https://prettier.io/docs/en/configuration.html#configuration-overrides) to your `.prettierrc`.

Since the linked page has everything a user should need.

zemse:
Yea, there were repetitions and I wished to change it into something but didn't know what. Thanks for the link, updating this right away.

PR #333, opened 2020-09-03
Print returns in function types at the end

### LINKED ISSUE #325: Function type declaration are rewritten incorrectly
**The original code:**
```
contract Broken {
    struct Something {
        // This gives an error
        function (address) external view returns (uint256) getSomething;
    }
}
```
**The rewrite:**
```
contract Broken {
    struct Something {
        // This gives an error
        function(address) returns (uint256) external view getSomething;
    }
}
```

### LINKED ISSUE #330: The plugin is incorrectly changing the position of function argument's modifiers
If I write a function that takes another function as a parameter, the plugin changes the position of modifiers applied to that function passed as a parameter.

For example, if I write:
```
function reduce(
        uint256[] memory self, 
        function(uint256, uint256) pure returns (uint256) f
    ) internal pure returns (uint256 r) {
        r = self[0];
        for (uint256 i = 1; i < self.length; i++) {
            r = f(r, self[i]);
        }
    }
```
The plugin changes it to:
```
function reduce(
        uint256[] memory self,
        function(uint256, uint256) returns (uint256) pure f
    ) internal pure returns (uint256 r) {
        r = self[0];
        for (uint256 i = 1; i < self.length; i++) {
            r = f(r, self[i]);
        }
    }
```
which is an invalid function signature.

The work-around that I found to work for now is to use the `// prettier-ignore` comment before that specific line:
```
function reduce(
        uint256[] memory self,
        // prettier-ignore
        function(uint256, uint256) pure returns (uint256) f
    ) internal pure returns (uint256 r) {
        r = self[0];
        for (uint256 i = 1; i < self.length; i++) {
            r = f(r, self[i]);
        }
    }
```

For completion, this is my current `.prettierrc`:
```
{
  "overrides": [{
    "files": "*.sol",
    "options": {
      "printWidth": 80,
      "tabWidth": 4,
      "useTabs": false,
      "singleQuote": false,
      "bracketSpacing": false,
      "explicitTypes": "always"
    }
  }]
}
```

fvictorio:
Fixes #325 
Fixes #330

### DIFF
--- a/tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
@@ -489,6 +489,22 @@ contract ConstructorWithoutVisibility {
     x = _x;
   }
 }
+
+contract FunctionTypes {
+    struct Something {
+        function (address) external view   returns (uint256) getSomething;
+    }
+
+function reduce(
+        uint256[] memory self, 
+        function  (uint256, uint256) pure   returns (uint256) f
+    ) internal pure returns (uint256 r) {
+        r = self[0];
+        for (uint256 i = 1; i < self.length; i++) {
+            r = f(r, self[i]);
+        }
+    }
+}
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 // Examples taken from the Solidity documentation online.
 
@@ -1025,6 +1041,22 @@ contract ConstructorWithoutVisibility {
 
     constructor(uint256 _x) {
         x = _x;
+    }
+}
+
+contract FunctionTypes {
+    struct Something {
+        function(address) external view returns (uint256) getSomething;
+    }
+
+    function reduce(
+        uint256[] memory self,
+        function(uint256, uint256) pure returns (uint256) f
+    ) internal pure returns (uint256 r) {
+        r = self[0];
+        for (uint256 i = 1; i < self.length; i++) {
+            r = f(r, self[i]);
+        }
     }
 }
 
--- a/Foo.sol
+++ b/Foo.sol
@@ -1,0 +1,15 @@
+contract FunctionTypes {
+    struct Something {
+        function (address) external view   returns (uint256) getSomething;
+    }
+
+function reduce(
+        uint256[] memory self, 
+        function  (uint256, uint256) pure   returns (uint256) f
+    ) internal pure returns (uint256 r) {
+        r = self[0];
+        for (uint256 i = 1; i < self.length; i++) {
+            r = f(r, self[i]);
+        }
+    }
+}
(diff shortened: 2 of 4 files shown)


tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1028,4 +1044,20 @@ contract ConstructorWithoutVisibility {
...
+        uint256[] memory self,
+        function(uint256, uint256) pure returns (uint256) f
+    ) internal pure returns (uint256 r) {
+        r = self[0];
+        for (uint256 i = 1; i < self.length; i++) {
+            r = f(r, self[i]);
+        }
+    }
+}
+
```
fvictorio, line 1062: @adjicf @andrekorol can you check if this looks good to you?

andrekorol: I've cloned the function-types branch and successfully ran the Jest tests locally.

But when running prettier with the function-types branch installed on node_modules/prettier-plugin-solidity, I still get the same ParserError: 
`contracts/Pyramid.sol[error] contracts/Pyramid.sol: ParserError: extraneous input 'pure' expecting {',', ')'} (10:44)`

I wonder if that's a problem with the Solidity parser ([federicobond/solidity-parser-antlr](https://github.com/federicobond/solidity-parser-antlr)) used by this plugin.

I noticed that the tests on AllSolidityFeatures.sol are using `pragma solidity ^0.4.0;`, but the contract I'm getting the ParserError was using `pragma solidity >=0.4.16 <0.8.0;`, and the version of solc I was compiling it on is 0.7.0. I'm not sure, but another possibility is that this error only arises when using solc-0.7.x.

This is nothing critical or urgent though, the contract in question is from the Solidity documentation and can be found in this section: [Function Types](https://github.com/ethereum/solidity/blob/53d497fc31a219242a80689bdee4a9f904f8ae2d/docs/types/value-types.rst#function-types).

mattiaerre: I'll merge as is and maybe we can have a follow-up PR?

mattiaerre (approved):
👍

tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
andrekorol: That's fine by me :)

I would have to take a closer look at the problem before creating a new PR, but from quick experimentation, the problem only seems to come up when using solc-0.7.

fvictorio: @andrekorol can you open an issue if that's still happening to you? And btw, if you think it's a bug with the parser (which it's likely), you can open an issue in https://github.com/solidity-parser/parser too

andrekorol: @fvictorio Sure! No problem :)

PR #349, opened 2020-10-01
Adjust stateMutability in function modifier order

benkostr:
Not sure if you want this, but it would follow the Solidity style guide.
https://solidity.readthedocs.io/en/v0.7.2/style-guide.html#function-declaration

Specifically:
The modifier order for a function should be:

1. Visibility
2. Mutability
3. Virtual
4. Override
5. Custom modifiers

### DIFF
--- a/src/nodes/FunctionDefinition.js
+++ b/src/nodes/FunctionDefinition.js
@@ -89,9 +89,9 @@ const FunctionDefinition = {
         group(
           concat([
             visibility(node),
+            stateMutability(node),
             virtual(node),
             override(node, path, print),
-            stateMutability(node),
             modifiers(node, path, print),
             returnParameters(node, path, print),
             signatureEnd(node)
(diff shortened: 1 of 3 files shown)


fvictorio:
I'm personally ok with this. But I'm surprised that tests passed, this means that we need more tests for the modifiers order. Would you mind adding some? We don't need to try all possible orders, that would be crazy, but just a few.

fvictorio:
For the record: OpenZeppelin seems to use this order

mattiaerre:
thanks for this PR @benkostr 🙏 
are we ok w/ this drop in coverage @fvictorio ?
it seems weird to me too that "just" flipping a line has that impact

fvictorio (approved):

fvictorio:
Yeah, I'm pretty sure it's a false negative.

Issue #374, opened 2020-11-03
bracketSpacing and JS files


yhuard:
Hi!

Thank you for this great plugin! :)

I'd like to report an issue with: https://github.com/prettier-solidity/prettier-plugin-solidity/blob/c631cf86a00117a89ef73d9946880e38d996faa7/src/options.js#L10

Prettier's [default value](https://github.com/prettier/prettier/blob/5768689995b39c446a2fad327a5a422926d500dd/src/common/common-options.js#L11) is `true`, and because the solidity plugin sets it to `false`, our JS files have been modified...

FYI, in our project, we use Prettier's default config (no config file) and we've only added the Solidity plugin. Ideally, adding the Solidity plugin should not impact other files than `.sol`.

By the way, I'm wondering, is it necessary to redefine the `bracketSpacing` option in the plugin? Other plugins like [plugin-php](https://github.com/prettier/plugin-php/blob/master/src/options.js) or [plugin-xml](https://github.com/prettier/plugin-xml/blob/master/src/plugin.js) only define new options, they don't redefine existing Prettier options. 🤔 

Thank you!

fvictorio:
Thanks @yhuard, you are absolutely right!

[Here's](https://prettier.io/docs/en/plugins.html#options) the relevant documentation (that section and the next one).

We should stop defining options that already exist in prettier core, and we should use `defaultOptions` instead (otherwise, if we just remove the options, everyone using the plugin will suddenly have their code reformatted in the next version).

This seems like an easy change if you feel like sending a PR!

yhuard:
Hi @fvictorio, sure, I'll create a PR when I have some free time. 👌

PR #375, opened 2020-11-03
Avoid redefining existing Prettier options

yhuard:
Fixes https://github.com/prettier-solidity/prettier-plugin-solidity/issues/374

The tests pass, `defaultOptions` looks good, all that was required was to remove the existing Prettier options.


fvictorio (approved):
That was fast, LGTM!

PR #392, opened 2020-11-15
Use solidity-comments-extractor

### LINKED ISSUE #355: Issue when parsing files with numbers like 100_000
There is an issue parsing numbers with underscores, for example 100_000.
To repro:
```
pragma solidity 0.6.12;

contract Bug {
  // This is a comment
  uint public hello;

  // Another comment
  uint public bigNum = 100_000;

  // This will disappear
  uint public magic;
}
```

Output:
```
pragma solidity 0.6.12;

contract Bug {
    // This is a comment
    uint256 public hello;

    // Another comment
    uint256 public bigNum = 100_000;

    uint256 public magic;
}
```

Comments will be removed after the `_` and as you can see `  // This will disappear` went away.

Versions:
```
"prettier": "^2.1.2",
"prettier-plugin-solidity": "^1.0.0-alpha.58"
```

fvictorio:
Closes #355.

This PR replaces the `extract-comments` package with `solidity-comments-extractor`, that I wrote (you can check it out [here](https://github.com/fvictorio/solidity-comments-extractor), but beware: the code is awful).

This not only closes the linked issue, it should also be more robust, since using a javascript parser to extract comments in solidity is bound to give us more issues in the future.

Plus, I think it should make the plugin faster. Maybe. I'm not sure.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7972,9 +7972,9 @@
       }
     },
     "solidity-comments-extractor": {
-      "version": "0.0.2",
-      "resolved": "https://registry.npmjs.org/solidity-comments-extractor/-/solidity-comments-extractor-0.0.2.tgz",
-      "integrity": "sha512-8twZSUh6oHpHrdq2DxXTd646Vjh3V7SwLGL91UEnBFM5ru4RxnaH9A536VUdVG+u/y9W2WyFNE62Pr+8Qnk4uA=="
+      "version": "0.0.4",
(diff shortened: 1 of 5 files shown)


mattiaerre (approved):
this is great thanks @fvictorio happy if you wanna move the package to the prettier-solidty organization. either way, I approve 👍

fvictorio:
Yes, let's do that!

PR #393, opened 2020-11-19
Don't join multi-part literals

### LINKED ISSUE #391: [Suggestion] Don't join string literals that are split
Edit: (Related: https://github.com/solidity-parser/parser/issues/32 this is required to allow hex literals to have the same formatting)

I have a 148 character long string that is hard to look at in the source code. So I've split it using multi-line and prettier-ignore

Similar to JavaScript prettier, if I write a string as `'abc' + 'def'` than it tends to format things nicer than when I keep everything as one long string, the format process feels free to play around with my "chunks" and their position to make the "prettiest" format.

Currently split strings get joined.

Example:

```solidity
string memory a = 'abc'
                    'def'
          'ghi';
```

Current Result:

```solidity
string memory a = 'abcdefghi';
```

Suggestion:

```solidity
string memory a = 'abc'
    'def'
    'ghi';
```

Or maybe have them be same line unless going over a certain length:

```solidity
// assume including the ghi chunk would put it over the one line length limit.
string memory a = 'abcdef'
    'ghi';
// OR
string memory a = 'abc' 'def'
    'ghi';
```

fvictorio:
Closes #391

This PR prevents prettier from joining multi-part strings and hex literals. It also prints each one in a separate line when the line is too long.

### DIFF
--- a/tests/MultipartStrings/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/MultipartStrings/__snapshots__/jsfmt.spec.js.snap
@@ -1,0 +1,30 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`MultipartStrings.sol 1`] = `
+contract MultipartStrings {
+  bytes b1 = hex'beef';
+  bytes b2 = hex"beef";
+  bytes b3 = hex"beef" hex"c0ffee";
+  bytes b4 = hex"beeeeeeeeeeeeeeeeeeeeeef" hex"c0000000000ffeeeeeeeeeeeeeeeeeee";
+
+  string s1 = "foo";
+  string s2 = "foo" "bar";
+  string s3 = "foofoofoofooofoofoofofoooofofoo" "barbarbrabrbarbarbabrabrbabr";
+}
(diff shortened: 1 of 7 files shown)


tests/MultipartStrings/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -0,0 +1,30 @@
...
+    bytes b4 =
+        hex"beeeeeeeeeeeeeeeeeeeeeef"
+        hex"c0000000000ffeeeeeeeeeeeeeeeeeee";
+
+    string s1 = "foo";
+    string s2 = "foo" "bar";
+    string s3 =
+        "foofoofoofooofoofoofofoooofofoo"
+        "barbarbrabrbarbarbabrabrbabr";
+}
```
mattiaerre, line 28: very nice 👍

mattiaerre (approved):
👍

junderw:
LGTM!!! Can't wait to clean up some contracts!

junderw:
Also in semi-related issues to allow for multiline hex literals output by this new formatter:

https://github.com/sc-forks/solidity-coverage/issues/575

https://github.com/protofire/solhint/issues/273

fvictorio:
@junderw they just need to update the parser to the latest version :)

PR #394, opened 2020-11-19
Update README.md

fvictorio:
Add logo and twitter badge. Change wording in intro. Change the basic usage instructions.


README.md
```diff
@@ -16,16 +21,22 @@ Install both `prettier` and `prettier-plugin-solidity`:
 npm install --save-dev prettier prettier-plugin-solidity
 ```
 
+Run prettier in your contracts:
+
+```
+npx prettier --write 'contracts/**/*.sol'
```
mattiaerre, line 27: this still requires to install the plugin locally; I'm not sure I would mention this in the readme; what do you think @fvictorio ?

fvictorio: It's meant as a second step after installing. Maybe I wasn't clear about that, but it works fine and it's (I think) the easiest way to try it out quickly after installing it.

PR #407, opened 2020-12-09
Upgrade solidity parser

fvictorio:
The parser is now bundled, because the antlr dependency requires node 14
and that could cause issues with yarn (although I don't quite understand
why yarn sometimes complains and sometimes it doesn't).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1592,12 +1592,9 @@
       }
     },
     "@solidity-parser/parser": {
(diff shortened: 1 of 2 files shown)


mattiaerre (approved):
LGTM, I guess we also need to release another version because of this?

fvictorio:
> LGTM, I guess we also need to release another version because of this?

Yep!

PR #428, opened 2021-02-05
Ignore comments in an ignored block

Janther:
if any node has the `ignore-prettier` comment, we will iterate through the whole object and mark the comments as printed.

I also took the liberty to update the dependencies and remove most listing errors in the test engine that we had.

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,22 +61,22 @@
     "node": ">=10"
   },
   "devDependencies": {
-    "codecov": "^3.6.5",
-    "eslint": "^7.3.1",
-    "eslint-config-airbnb-base": "^14.1.0",
+    "codecov": "^3.8.1",
+    "eslint": "^7.19.0",
+    "eslint-config-airbnb-base": "^14.2.1",
     "eslint-config-prettier": "^7.0.0",
-    "eslint-plugin-import": "^2.20.2",
-    "jest": "^26.0.1",
+    "eslint-plugin-import": "^2.22.1",
+    "jest": "^26.6.3",
     "jest-mock-now": "^1.3.0",
-    "jest-watch-typeahead": "^0.6.0"
+    "jest-watch-typeahead": "^0.6.1"
   },
   "dependencies": {
     "@solidity-parser/parser": "^0.11.0",
     "dir-to-object": "^2.0.0",
-    "emoji-regex": "^9.0.0",
+    "emoji-regex": "^9.2.1",
     "escape-string-regexp": "^4.0.0",
-    "prettier": "^2.0.5",
-    "semver": "^7.3.2",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -61,22 +61,22 @@
...
     "dir-to-object": "^2.0.0",
-    "emoji-regex": "^9.0.0",
+    "emoji-regex": "^9.2.1",
     "escape-string-regexp": "^4.0.0",
-    "prettier": "^2.0.5",
-    "semver": "^7.3.2",
+    "prettier": "^2.2.1",
+    "semver": "^7.3.4",
     "solidity-comments-extractor": "^0.0.4",
     "string-width": "^4.2.0"
```
mattiaerre, line 81: did you have to update all these deps yourself? have they been missed by dependabot? really?

tests_config/helpers.js
```diff
@@ -0,0 +1,44 @@
...
+
+function read(filename) {
+  return fs.readFileSync(filename, 'utf8');
+}
+
+function parse(string, opts) {
+  return prettier.__debug.parse(string, opts, /* massage */ true).ast;
+}
+
+module.exports = { mergeDefaultOptions, parse, prettyprint, raw, read };
```
mattiaerre, line 44: good idea to move them our into a separate module 👍

tests_config/run_spec.js
```diff
@@ -73,45 +76,4 @@ function run_spec(dirname, options) {
...
-function mergeDefaultOptions(parserConfig) {
-  return Object.assign(
-    {
-      plugins: [path.dirname(__dirname)],
-      printWidth: 80
-    },
-    parserConfig
-  );
-}
+global.run_spec = runSpec;
```
mattiaerre, line 79: nice refactoring

tests_config/run_spec.js
```diff
@@ -1,16 +1,19 @@
...
-
 const fs = require('fs');
 const path = require('path');
+const {
+  read,
+  mergeDefaultOptions,
+  prettyprint,
+  raw,
+  parse
+} = require('./helpers');
```
mattiaerre, line 10: I'd also alpha import here but I'm also ok to do it in a next PR

mattiaerre (approved):
that's fantastic, thanks @Janther

package.json
Janther: yeah dependabot does the heavy lifting for critical dependencies I believe. also I did not have any error pop up when I created the new tests and only started failing when I updated the dependancies. maybe the latest prettier did the trick.

mattiaerre: splendid, I'll merge and deploy thanks @Janther

PR #433, opened 2021-02-08
We use Prettier's FastPath strategy to avoid modifying parameters.

Janther:
This way we get rid of having to get rid of eslint's rule `no-param-reassign`

### DIFF
--- a/src/comments/ignore.js
+++ b/src/comments/ignore.js
@@ -1,16 +1,31 @@
-function ignoreComments(node) {
-  if (node === undefined || node === null) return;
+function ignoreComments(path) {
+  const node = path.getValue();
+  // We ignore anything that is not an object
+  if (node === null || typeof node !== 'object') return;
+
   const keys = Object.keys(node);
   keys.forEach((key) => {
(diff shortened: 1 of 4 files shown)


mattiaerre (approved):
thanks @Janther 👍

src/comments/ignore.js
```diff
@@ -1,16 +1,31 @@
-function ignoreComments(node) {
-  if (node === undefined || node === null) return;
+function ignoreComments(path) {
+  const node = path.getValue();
+  // We ignore anything that is not an object
+  if (node === null || typeof node !== 'object') return;
```
mattiaerre, line 4: do we wanna keep `node === undefined ||` as well?

mattiaerre (commented):
added an additional comment

src/comments/ignore.js
Janther: It's not necessary because of the object type check.
```Javascript
typeof undefined !== 'object' // true

typeof null !== 'object' // false
```

PR #448, opened 2021-03-14
Updating Tests to latest Prettier tests

Janther:
there are a few benefits of updating our testing framework.
  - the new snapshots give more details.
  - the `FULL_TEST` not only compares `AST` but also different `EOL`, `BOM`, and compares the file formatted twice (as described in Prettier's [documentation](https://github.com/prettier/prettier/blob/main/CONTRIBUTING.md#deeper-testing)) which led to the discovery with Assembly Labels adding an extra line if there is an existing one.


tests/EnumDefinitions/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1,13 +1,20 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`EnumDefinitions.sol 1`] = `
+exports[`EnumDefinitions.sol - {"bracketSpacing":true} format 1`] = `
```
Janther, line 3: things like these are very nice when reviewing a snapshot with specific `options`

tests/ExplicitVariableTypes/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -31,61 +37,62 @@ contract VariableTypesMixed {
...
+    event Event(uint _a, int _b, byte _c, uint _e, int _f, byte _g);
 
     function func(
-        uint256 _a,
-        int256 _b,
-        bytes1 _c,
-        uint256 _e,
-        int256 _f,
-        bytes1 _g
+        uint _a,
```
Janther, line 64: These changed because the order of the snapshots changed.

tests_config/run_spec.js
```diff
@@ -1,79 +1,389 @@
...
-        .replace(/\r\n/g, '\n')
-        .replace('<<<PRETTIER_RANGE_START>>>', (match, offset) => {
-          rangeStart = offset;
-          return '';
-        })
-        .replace('<<<PRETTIER_RANGE_END>>>', (match, offset) => {
-          rangeEnd = offset;
-          return '';
-        });
+// TODO: add support for standalone
```
Janther, line 3: I don't know if we need this but in the future we can add it.

tests_config/run_spec.js
```diff
@@ -1,79 +1,389 @@
...
-          raw(`${source + '~'.repeat(mergedOptions.printWidth)}\n${output}`)
-        ).toMatchSnapshot();
-      });
+const CURSOR_PLACEHOLDER = '<|>';
+const RANGE_START_PLACEHOLDER = '<<<PRETTIER_RANGE_START>>>';
+const RANGE_END_PLACEHOLDER = '<<<PRETTIER_RANGE_END>>>';
+
+// TODO: these test files need fix
+const unstableTests = new Map(
+  ['Assembly/Assembly.sol'].map((fixture) => {
```
Janther, line 20: Until we fix this Issue we won't compare this file formatted twice.

tests_config/run_spec.js
Janther: Felt bored and just added this :)

mattiaerre:
happy to approve this if you can fix the conflicts @Janther thanks

Janther:
AFK right now but the conflict might only be with the lock file since the merge of the eslint update

mattiaerre:
thanks @Janther this is a 104 files long PR, anything specific you'd like us to take a look at?

Janther:
Every `jsfmt.spec.js` in the test folder had the parser included in the call to `run_spec`.
Also every snapshot changed slightly because the new snapshots include more info.
So just have a look at a few of these examples in the `test` folder.

The files in the `test_config` folder are taken from Prettier’s repo. Only the `require_standalone.js` and `run_spec.js` were altered.
  * `require_standalone.js` requires the `standalone` module instead of parsing it and executing it in a separate empty context .
  * `run_spec.js` has some variables and checks, that only dealt with specific languages and parsers supported directly by Prettier, removed.

Probably would suggest to only check these 2 files.

Finally there are some changes in the Jest config mainly taken from Prettier’s repo. And a few changes in the `package.json` scripts.

`AST_COMPARE` doesn't exist anymore, it was replaced by `FULL_TEST` which does much more now.
Running the tests won’t collect coverage unless we run them with the `FULL_TEST` flag. this way we won't get noisy feedback when we run a specific test like `npm test tests/Assembly`.

Finally you could have a look and play with the result.

mattiaerre:
thanks for the recap @Janther this is great

mattiaerre:
happy to merge if @fvictorio is good w/ that and once the conflict is resolved 😅

package.json
```diff
@@ -5,13 +5,13 @@
...
-    "eslint": "eslint --ignore-pattern 'src/prettier-comments/**' 'src/**'",
+    "eslint": "eslint 'src/**'",
     "generateindexes": "SCRIPT=true node scripts/generateIndexes.js",
     "lint": "npm run eslint && npm run prettier -- --list-different",
     "lint:fix": "npm run eslint -- --fix && npm run prettier -- --write",
-    "prettier": "prettier '{.,{src,tests_config}/**}/*.{js{,on},md}' 'tests/**/jsfmt.spec.js' '!src/prettier-comments/**/*'",
+    "prettier": "prettier '{.,{src,tests_config}/**}/*.{js{,on},md}' 'tests/**/jsfmt.spec.js'",
     "test": "jest",
-    "test:all": "AST_COMPARE=true npm t -- --coverage"
+    "test:all": "FULL_TEST=1 jest"
```
mattiaerre, line 14: can we use `true` instead of `1`? thanks

jest.config.js
```diff
@@ -1,5 +1,7 @@
+const { FULL_TEST } = process.env;
+
 module.exports = {
-  collectCoverage: true,
+  collectCoverage: !!FULL_TEST,
```
mattiaerre, line 4: related to my other comment below; not a huge fan of the `!!` operator

tests_config/require_standalone.js
```diff
@@ -0,0 +1,22 @@
+const prettier = require('prettier/standalone');
+const solidityPlugin = require('../src/index');
+
+module.exports = {
+  formatWithCursor(input, options) {
+    const $$$options = {
```
mattiaerre, line 6: `$$$options` can you tell me more about it?

tests_config/setup.js
```diff
@@ -0,0 +1,7 @@
+const runSpec = require('./run_spec.js');
```
mattiaerre, line 1: interesting, it looks like from this file on GH is no longer formatting js? 🤷‍♂️

mattiaerre (changes requested):

tests_config/require_standalone.js
Janther: Just a name to be different from `options` it came from the original file. That one instead of importing the code, it reads the files and treats it as a string to be evaluated by the `vm`. I didn't have much time to go through this process so there's some room for improvement here.

package.json
Janther: I used `1` as it is what's used in Prettier's scripts. It can be anything though as the variable is a string and the code just checks for a `truthy` value. It would also run full tests with `FULL_TEST=false`.

README.md
```diff
@@ -152,8 +152,9 @@ If you want a different configuration for your javascript and solidity files, yo
 1. [Fork it](https://github.com/prettier-solidity/prettier-plugin-solidity/fork)
 2. Create your feature branch (`git checkout -b feature/fooBar`)
 3. Commit your changes (`git commit -am 'Add some fooBar'`)
-4. Push to the branch (`git push origin feature/fooBar`)
-5. Create a new Pull Request
+4. All existing test and coverage must pass (`npm run test:all`), if coverage drops below 100% add missing tests.
```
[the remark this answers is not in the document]
Janther: Not sure what else can I write here. Perhaps I'm too close to the forest to see the trees. Maybe you can add something here?

jest.config.js
Janther: Solved

tests_config/setup.js
Janther: Looks normal in my screen.

README.md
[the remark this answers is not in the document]
mattiaerre: this is my bad sorry, the desc is perfect

package.json
mattiaerre: alright, I'm ok to keep 1 then thanks for the explanation

mattiaerre (approved):
thanks @Janther good job 🙏

Issue #451, opened 2021-03-16
prettier-plugin-solidity fails on ethereum-optimism/contracts


janek26:
If you try to use the Plugin to format [this file](https://github.com/ethereum-optimism/contracts/blob/77d30d66b93fa65782d2f15d07843822b28df9e6/contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol) you get the following error thrown:

```
contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol
[error] contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol: Error: Comment "uint40 _shouldStartAtElement," was not printed. Please report this error!
[error]     at /Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/index.js:14350:13
[error]     at Array.forEach (<anonymous>)
[error]     at Object.ensureAllCommentsPrinted (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/index.js:14348:15)
[error]     at coreFormat (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/index.js:14912:12)
[error]     at format (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/index.js:15131:14)
[error]     at Object.formatWithCursor (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/index.js:57542:12)
[error]     at format (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/bin-prettier.js:15883:14)
[error]     at Object.formatFiles (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/bin-prettier.js:16067:16)
[error]     at Object.run (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/bin-prettier.js:16538:14)
[error]     at Object.<anonymous> (/Users/janek.rahrt/Projects/optimism/contracts/node_modules/prettier/bin-prettier.js:16553:5)
```

It seems to be due to [Line 213](https://github.com/ethereum-optimism/contracts/blob/77d30d66b93fa65782d2f15d07843822b28df9e6/contracts/optimistic-ethereum/iOVM/chain/iOVM_CanonicalTransactionChain.sol#L213)

fvictorio:
[...]

PR #452, opened 2021-03-17
Comments for an empty parameter list

Janther:
solves #451

### DIFF
--- a/tests/Comments/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/Comments/__snapshots__/jsfmt.spec.js.snap
@@ -86,7 +86,16 @@ contract Comments9 {
     }
   }
 }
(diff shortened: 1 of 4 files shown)


mattiaerre (approved):
thanks @Janther can we merge this before the other outstanding PRs?

Janther:
Yes.
The test PR rewrites all snapshots. Since this one added a change in a snapshot there will be a conflict but running the tests should solve it

tests/Comments/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -194,4 +203,16 @@ contract Comments9 {
     }
 }
 
+interface Comments10 {
+    function someFunction(
+        // the first value
+        // the second value
+        // the lats value
+    )/* comment outside the parameters */ external;
```
fvictorio, line 211: Is this correct? I think an space after the `)` would be better.

Issue #449, opened 2021-03-14
New lines before Assembly labels are not predictable


Janther:
This input:
```Solidity
function assemblyLabels() {
    assembly {
        let n := calldataload(4)
        let a := 1
        let b := a
    loop:
        jumpi(loopend, eq(n, 0))
        a add swap1
        n := sub(n, 1)
        jump(loop)

    loopend:
        mstore(0, a)
        return(0, 0x20)
}
```

will generate this output:
```Solidity
function assemblyLabels() {
    assembly {
        let n := calldataload(4)
        let a := 1
        let b := a

    loop:
        jumpi(loopend, eq(n, 0))
        a add swap1
        n := sub(n, 1)
        jump(loop)


    loopend:
        mstore(0, a)
        return(0, 0x20)
}
```
The AST are identical but there is an extra newline before `loopend`.

This was discovered in #448

fvictorio:
This will be fun -_-

Janther:
Already have a solution but am waiting for the new testing PR to be reviewed and approved 😉

fvictorio:
Oh, nice. I'll try to take a look asap!

PR #456, opened 2021-03-29
We avoid prepending a hardline before a LabelDefinition

Janther:
`LabelDefinition` uses a dedented first line.
This caused a side effect where all Labels would have an empty line before them, this looked a bit quirky but there wasn't anything wrong with it.

After adding the new in depth testing, we recognised that, on the second format, this empty line would be registered as an empty line added by the developer to space out code, and we would preserve this line in `printPreservingEmptyLines`, and we would end up with 2 empty lines instead of one.

This fix might look a bit hacky but it's an edge case given the particular way of printing assembly labels.

### DIFF
--- a/tests/Assembly/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/Assembly/__snapshots__/jsfmt.spec.js.snap
@@ -125,6 +125,7 @@ loop:
     a add swap1
     n := sub(n, 1)
     jump(loop)
+
 loopend:
     mstore(0, a)
     return(0, 0x20)
@@ -322,7 +323,6 @@ contract Assembly {
             let n := calldataload(4)
             let a := 1
             let b := a
-
         loop:
             jumpi(loopend, eq(n, 0))
             a
@@ -339,7 +339,6 @@ contract Assembly {
         assembly {
             let x := 8
             jump(two)
-
         one:
             // Here the stack height is 2 (because we pushed x and 7),
             // but the assembler thinks it is 1 because it reads
@@ -347,11 +346,9 @@ contract Assembly {
             // Accessing the stack variable x here will lead to errors.
             x := 9
             jump(three)
-
         two:
             7 // push something onto the stack
             jump(one)
-
         three:
         }
     }
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -18,9 +18,9 @@ const CURSOR_PLACEHOLDER = '<|>';
 const RANGE_START_PLACEHOLDER = '<<<PRETTIER_RANGE_START>>>';
 const RANGE_END_PLACEHOLDER = '<<<PRETTIER_RANGE_END>>>';
 
-// TODO: these test files need fix
+// Here we add files that will not be the same when formating a second time.
 const unstableTests = new Map(
-  ['Assembly/Assembly.sol'].map((fixture) => {
+  [].map((fixture) => {
     const [file, isUnstable = () => true] = Array.isArray(fixture)
       ? fixture
       : [fixture];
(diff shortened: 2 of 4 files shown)


tests/Assembly/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -125,6 +125,7 @@ loop:
     a add swap1
     n := sub(n, 1)
     jump(loop)
+
```
fvictorio, line 128: Why does this have a space but the `one:`, `two:`, etc. labels below don't?

tests_config/run_spec.js
```diff
@@ -18,17 +18,25 @@ const CURSOR_PLACEHOLDER = '<|>';
 const RANGE_START_PLACEHOLDER = '<<<PRETTIER_RANGE_START>>>';
 const RANGE_END_PLACEHOLDER = '<<<PRETTIER_RANGE_END>>>';
 
-// TODO: these test files need fix
+// Here we add files that will not be the same when formating a second time.
 const unstableTests = new Map(
-  ['Assembly/Assembly.sol'].map((fixture) => {
+  [].map((fixture) => {
```
mattiaerre, line 23: noice now this is empty 😊

tests_config/run_spec.js
```diff
@@ -18,17 +18,25 @@ const CURSOR_PLACEHOLDER = '<|>';
...
       ? fixture
       : [fixture];
     return [path.join(__dirname, '../tests/', file), isUnstable];
   })
 );
 
-const unstableAstTests = new Map();
+// Here we add files that will not have the same AST after being formated.
+const unstableAstTests = new Map(
+  [].map((fixture) => {
```
mattiaerre, line 33: same as above

tests/Assembly/__snapshots__/jsfmt.spec.js.snap
mattiaerre: yeah, they actually got removed

Janther: I added a space here just to have the same code as the example in #449
That way we have an example of no empty line before and 1 empty line before

tests_config/run_spec.js
Janther: I decided to add the function within since an empty `new Map();` doesn't really explain how this map should be implemented in case we had an example where ASTs are changed. 
We don't have inconsistent AST but we do sometimes change it in the parsing step.
Stuff like changing `uint` to `uint256` and enforcing parentheses for easy to read code. For more correctness we could also add a new test that compares the compiled code of contracts where we know that the AST changes.
Until now I have used this [project](https://github.com/Janther/prettier-ast-changing-proposals) to prove that those AST changes don't affect the compiled code, but would be a nice addition to our process and we could keep track on new compiler versions that might affect this.

tests/Assembly/__snapshots__/jsfmt.spec.js.snap
Janther: > yeah, they actually got removed

yeah, before this PR we were always printing an extra line. In the other snapshots, those lines got removed in the output.
This line is from the input of `tests/Assembly/Assembly.sol` so it is added.

mattiaerre (approved):
good stuff @Janther

tests/Assembly/__snapshots__/jsfmt.spec.js.snap
fvictorio: Oh, sorry :man_facepalming: I shouldn't review code when I'm too tired.

fvictorio (approved):

PR #459, opened 2021-03-31
Breaking changes in v0.8.0

Janther:
v0.8.0 added [breaking changes](https://docs.soliditylang.org/en/v0.8.0/080-breaking-changes.html#silent-changes-of-the-semantics) that conflicted with some desicions we made while formatting.

The ones affecting us are mainly:
  - The type `byte` has been removed. It was an alias of `bytes1`.
  - Exponentiation is right associative, i.e., the expression `a**b**c` is parsed as `a**(b**c)`. Before 0.8.0, it was parsed as `(a**b)**c`.

Since we are not tracking the pragma of the file, we should simply avoid special formatting of these.

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -29,14 +29,18 @@ function parse(text, parsers, options) {
         : `hex"${ctx.value.slice(4, -1)}"`;
     },
     ElementaryTypeName(ctx) {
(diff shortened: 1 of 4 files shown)


fvictorio (approved):

mattiaerre (approved):
🙏

PR #461, opened 2021-03-31
Using solc to compare bytecode on specific contracts

Janther:
There are 3 different cases where we make changes to the AST while parsing.

  1. `HexLiteral`: if the plugin needs to change the quote of a `HexLiteral` because of the `singleQuote` setting, this is better done while parsing to avoid throwing an error in the tests for a different AST after formatting.
  2. Types and Alias: `uint` and `int` are aliases for `uint256` and `int256` respectively. These will generate different ASTs when using a different setting for `explicitTypes`
  3. Parentheses: To help the reading of mathematic expressions, we enforce the use the use of Parentheses in specific cases.
    - ie: `a / b * c` will be formatted as `(a / b) * c`.
    This clearly implies a change in the AST.

As nice as these changes to the AST can be for formatting, the compiled bytecode should remain the same. To enforce this, this PR will add the ability to compare the bytecode of some contracts designed to trigger all of these AST changes.

### DIFF
--- a/tests_config/run_spec.js
+++ b/tests_config/run_spec.js
@@ -38,6 +38,35 @@ const unstableAstTests = new Map(
   })
 );
 
+const astChangedTests = new Map(
+  [
+    [
+      'ExplicitVariableTypes/ExplicitVariableTypes.sol',
+      (options) =>
+        options.explicitTypes === undefined ||
+        options.explicitTypes !== 'preserve'
+    ],
+    'Parentheses/AddNoParentheses.sol',
+    'Parentheses/SubNoParentheses.sol',
+    'Parentheses/MulNoParentheses.sol',
+    'Parentheses/DivNoParentheses.sol',
+    'Parentheses/ModNoParentheses.sol',
+    'Parentheses/ExpNoParentheses.sol',
+    'Parentheses/ShiftLNoParentheses.sol',
+    'Parentheses/ShiftRNoParentheses.sol',
+    'Parentheses/BitAndNoParentheses.sol',
+    'Parentheses/BitOrNoParentheses.sol',
+    'Parentheses/BitXorNoParentheses.sol',
+    'Parentheses/LogicNoParentheses.sol',
+    'HexLiteral/HexLiteral.sol'
+  ].map((fixture) => {
+    const [file, compareBytecode = () => true] = Array.isArray(fixture)
+      ? fixture
+      : [fixture];
+    return [path.join(__dirname, '../tests/', file), compareBytecode];
+  })
+);
+
 const isUnstable = (filename, options) => {
   const testFunction = unstableTests.get(filename);
 
@@ -50,6 +79,16 @@ const isUnstable = (filename, options) => {
 
 const isAstUnstable = (filename, options) => {
   const testFunction = unstableAstTests.get(filename);
+
+  if (!testFunction) {
+    return false;
+  }
+
+  return testFunction(options);
+};
+
+const shouldCompareBytecode = (filename, options) => {
+  const testFunction = astChangedTests.get(filename);
 
   if (!testFunction) {
     return false;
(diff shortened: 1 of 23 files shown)


tests_config/run_spec.js
```diff
@@ -305,6 +344,20 @@ function runTest({
...
   }
+  if (shouldCompareBytecode(filename, formatOptions)) {
+    test(`[${parser}] compare Bytecode`, () => {
+      // We require the compiler here as it makes all tests slow when added at
+      // the top of the file.
+      // TODO investigate this warning
+      //   - A worker process has failed to exit gracefully and has been force
+      //     exited. This is likely caused by tests leaking due to improper
+      //     teardown. Try running with --detectOpenHandles to find leaks.
+      const compileContract = require('./utils/compile-contract');
```
Janther, line 355: I'd appreciate if someone could investigate how to include the compiler, without incurring in greater testing times.

tests_config/utils/compile-contract.js
```diff
@@ -0,0 +1,34 @@
...
+    settings: {
+      outputSelection: {
+        '*': {
+          '*': ['*']
+        }
+      }
+    }
+  };
+  const output = JSON.parse(solc.compile(JSON.stringify(input)));
+  const compiledContracts = output.contracts[filename];
```
Janther, line 20: This script was made only considering the happy path.
Would be nice if someone could add cases where the compiler fails.

fvictorio: What would we check if the compiler fails? In that case there would be no output. We could check that the compiler fails in both cases, but I'm not sure if that's very useful.

Janther: Just added a commit dealing with this.
The test won't pass unless the developer provides a compilable contract for the test.
This way we'll catch things like compiler versions not up to date on the tests that require compiling.

tests_config/run_spec.js
fvictorio: There is a way to improve the time: download a native compiler and use that instead. It's significantly faster. But this means:

1. Downloading the correct compiler according to the OS (and falling back to the WASM one if the OS is not supported)
2. Avoiding a re-download if it was previously download
3. (Optional) Cache it in the CI

I can do this since we do something similar (even more involved) in Hardhat.

That being said, it takes 1m to do this, so I don't think it's worth it? I'd rather re-visit this in the future if it's a problem, but for now I would say that we just run `npm test` during development and use `npm run test:all` before pushing (or just checking the result of the CI).

tests_config/run_spec.js
```diff
@@ -38,6 +38,35 @@ const unstableAstTests = new Map(
   })
 );
 
+const astChangedTests = new Map(
```
fvictorio, line 41: Two questions here:

1. Is the name of this variable accurate?
2. Why is the bytecode comparison only done for these files? Why don't we do it for all of them?

fvictorio: Ohh, the bytecode comparison is *only* done for these files, I understand now. I still think the name of the variable is not very descriptive.

fvictorio:
I would love to run this bytecode comparison in all our test files. This is **a lot of work**, because:

* We need to make all our test files valid
* We need to support multiple compilers. This can be done either by adding a compiler downloader or installing multiple `solc` dependencies with different versions (there is a way to do this with npm)
* Since we support multiple versions, we'll need to split our files so that we cover different scenarios (for example, functions with and without visibility)

But I think this PR is already too big, and it covers good ground. I think we can merge it and create an issue to do the full thing another day.

tests_config/run_spec.js
Janther: 1. Just changed the name slightly to be more precise.
2. For now I believe AST comparison should be enough and byte-code comparison applies only on the cases when we choose to change the AST.
  I agree that best case would be to ultimately compare byte codes but I also see all of the work needed to do that properly.
  Making sure every test is compilable and also keeping track of major syntax changes across different compilers.

mattiaerre (approved):
👍

PR #462, opened 2021-04-02
group unchecked blocks

Janther:
Having a look at the `unchecked` statement in the wild, it's usually applied to a single statement and unlikely to be broken into multiple lines, so I decided to `group` it so it only breaks into multiple lines if the content of the block breaks.

### DIFF
--- a/tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
@@ -528,8 +528,16 @@ contract WithUncheckedBlock {
     unchecked   {   x--;   }
     return x;
   }
-}
+  
+  function g() public pure returns (uint) {
+    uint x = 0;
+    unchecked   {   x--; return x; }
+  }
 
+  function mul(uint256 x, uint256 y) internal pure returns (uint256) {
+    unchecked { return x * y; }
+  }
+}
 =====================================output=====================================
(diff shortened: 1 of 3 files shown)


tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1098,11 +1106,21 @@ uint256 constant MULTIPLIER = 2**EXPONENT;
 
 contract WithUncheckedBlock {
     function f() public pure returns (uint256) {
+        uint256 x = 0;
+        unchecked {x--;}
+        return x;
+    }
+
+    function g() public pure returns (uint256) {
         uint256 x = 0;
```
mattiaerre, line 1115: qq I'm noticing we are also changing the type of this variable, since when are we doing this?

mattiaerre (approved):
👍  feel free to merge @Janther

tests/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
Janther: Hey,
This is the `explicit-types` option. We have been doing this since `v1.0.0-alpha-27`

Janther: In #461 you can see that the bytecode doesn't change, even if the AST slightly changes.

PR #466, opened 2021-04-06
keeping prettier-comments files as close to the source

Janther:
Just moving back the responsibility of printing the `unicode` keyword to the solidity printer and not to the files taken from the prettier utils.


mattiaerre (approved):
great a "real" refactoring w/ no test changes 💪

PR #478, opened 2021-04-19
Ignore auxdata at compile time

Janther:
thanks to @sambacha for reporting this.


mattiaerre (approved):
👍

mattiaerre:
it looks like we do not have any issue that this PR will close; thanks for taking care of it @Janther

Janther:
> it looks like we do not have any issue that this PR will close; thanks for taking care of it @Janther 
> 
> 

Yeah, it was mentioned in the telegram channel

PR #499, opened 2021-05-14
solidity-parser

mattiaerre:
# Description

- update solidity-parser

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,23 +1,23 @@
 version: 2
 updates:
-- package-ecosystem: npm
-  directory: "/"
-  schedule:
-    interval: daily
-  open-pull-requests-limit: 10
-  ignore:
-  - dependency-name: semver
-    versions:
-    - 7.3.5
-  - dependency-name: eslint
-    versions:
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -1,23 +1,23 @@
...
+    ignore:
+      - dependency-name: semver
+        versions:
+          - 7.3.5
+      - dependency-name: eslint
+        versions:
+          - 7.23.0
+      - dependency-name: '@solidity-parser/parser'
+        versions:
+          - 0.12.2
```
mattiaerre, line 17: do we still need this file btw? I think we don't

package.json
```diff
@@ -75,7 +75,7 @@
     "solc": "^0.8.3"
   },
   "dependencies": {
-    "@solidity-parser/parser": "^0.12.1",
+    "@solidity-parser/parser": "^0.12.2",
```
mattiaerre, line 78: the package.json already had v0.12.2 btw

Janther: yeah, dependabot only changes the lock files if it can help it

PR #502, opened 2021-05-21
Add support for custom errors

### LINKED ISSUE #485: Add support for error definition
https://docs.soliditylang.org/en/v0.8.4/control-structures.html?highlight=revert#error-handling-assert-require-revert-and-exceptions

Solidity now supports the creation of errors, to add syntax for stuff like:
```solidity
error Unauthorized();
if (!authorized()) revert Unauthorized()
```

Prettier throws:
`Error: mismatched input '(' expecting {';', '='}` on the `error Unauthorized()` line.

fvictorio:
Closes #485. Way overdue :sweat_smile:


fvictorio:
The failing tests make no sense to me. I know I can fix them by running `jest -u`, but if we need to make that every time we add a node, maybe these tests aren't adding any value?

Plus, if something is really missing in those files, the "real" tests will fail anyway.

@Janther @mattiaerre do you mind if I remove those tests?

fvictorio:
Ok, I preemptively removed them. I also excluded the `scripts` directory from the coverage check (which makes sense to me). But I'm totally fine with reverting that commit if you think it's a good idea to keep those tests.

package.json
```diff
@@ -75,7 +75,7 @@
     "solc": "^0.8.3"
   },
   "dependencies": {
-    "@solidity-parser/parser": "^0.12.2",
+    "@solidity-parser/parser": "^0.13.0",
```
Janther, line 78: ❤️

src/nodes/CustomErrorDefinition.js
```diff
@@ -0,0 +1,19 @@
...
+const printSeparatedList = require('./print-separated-list');
+
+const parameters = (node, path, print) =>
+  node.parameters && node.parameters.length > 0
+    ? printSeparatedList(path.map(print, 'parameters'))
+    : '';
+
+const CustomErrorDefinition = {
+  print: ({ node, path, print }) =>
+    concat(['error ', node.name, '(', parameters(node, path, print), ')', ';'])
```
Janther, line 16: the last `')', ';'` could just be `');'`

tests/CustomErrors/CustomErrors.sol
```diff
@@ -0,0 +1,20 @@
+pragma solidity ^0.8.4;
+
+error
+  TopLevelCustomError();
+  error TopLevelCustomErrorWithArg(uint    x)  ;
+error TopLevelCustomErrorArgWithoutName  (string);
+
+contract CustomErrors {
+  error
+    TopLevelCustomError();
```
Janther, line 10: this should be named `ContractCustomError`

tests/CustomErrors/CustomErrors.sol
```diff
@@ -0,0 +1,20 @@
...
+    error TopLevelCustomErrorWithArg(uint    x)  ;
+  error TopLevelCustomErrorArgWithoutName  (string);
+
+  function throwCustomError() {
+    revert
+      ContractCustomError();
```
Janther, line 16: This should be named `FunctionCustomError`

Janther:
The problem with snapshots in testing is that the snapshots need to be understandable and auditable. A extremely big list or a massive array are not something that snapshots were made for.
I agree that we need to test the scripts but would make much more sense to create a sample folder with just a few files and make the test generate the snapshot for that folder. That would be more auditable.

src/nodes/CustomErrorDefinition.js
fvictorio: Oops, good catch, that happened because I copied and modified the `EventDefinition` printer :sweat_smile:

fvictorio:
> The problem with snapshots in testing is that the snapshots need to be understandable and auditable. A extremely big list or a massive array are not something that snapshots were made for.
> I agree that we need to test the scripts but would make much more sense to create a sample folder with just a few files and make the test generate the snapshot for that folder. That would be more auditable.

Do you mind if we do/discuss that in another PR so we can get this merged?

(FWIW, I'm OK with not testing code-generating scripts if the generated code is tested, and we already have a 100% threshold in the directory where the generated code is added)

fvictorio:
Feedback applied, thank you @Janther!

Issue #511, opened 2021-05-27
Unchecked blocks and bracketSpacing


frangio:
Related to https://github.com/prettier-solidity/prettier-plugin-solidity/pull/462

Single-line `unchecked` blocks are rendered with bracket spacing or not according to the `bracketSpacing` option:

```
unchecked { counter._value += 1; }
```

```
unchecked {counter._value += 1;}
```

The `bracketSpacing` option is documented in Prettier as: "Print spaces between brackets in object literals". In this case these are not object literals but blocks. Prettier for JS never prints blocks in a single line, so we don't have a reference point for whether the option should be considered in this case.

My opinion on this is that block braces should always have spacing around them. Looking at the [Solidity docs](https://docs.soliditylang.org/en/v0.8.1/control-structures.html), they seem to agree with this: `unchecked` braces are shown with spacing, while other instances of braces like in function call syntax (`c.foo{value: 1eth}()`) are not shown with spacing.

fvictorio:
I think I'd rather always do this:

```
unchecked {
  counter._value += 1;
}
```

but I also understand why this is a place where solidity is different from most languages (this is not the same as a single-line if).

Anyway: agree that if we keep it in one line, we should always use spaces around the braces.

frangio:
I personally would also prefer the format in the snippet you shared! It's what we're currently using in OpenZeppelin Contracts, and it seems to be what Prettier does for blocks in JavaScript.

I guess it comes down to how these things are decided for prettier-solidity in general.

PR #514, opened 2021-05-28
block always breaks

Janther:
deals with #511 but also want to keep the conversation open for having a block in a single line separated by spaces


fvictorio (approved):

fvictorio:
I'd say we merge this. We can handle the edge case of an `unchecked` increment inside a for loop in another PR.

PR #516, opened 2021-05-28
chaining function calls: fixing undesirable indents

Janther:
We indent only if the chain doesn't belong to a variable declaration or a binary operation

### DIFF
--- a/src/nodes/MemberAccess.js
+++ b/src/nodes/MemberAccess.js
@@ -16,11 +16,28 @@ const isBeginnigOfChain = (path) => {
   return true;
 };
 
+const shallIndent = (path) => {
+  let i = 0;
+  let elderNode = path.getParentNode(i);
+  while (elderNode) {
+    if (
+      elderNode.type === 'VariableDeclarationStatement' ||
+      elderNode.type === 'BinaryOperation'
+    )
+      return false;
+    i += 1;
+    elderNode = path.getParentNode(i);
+  }
+  return true;
+};
+
 const MemberAccess = {
   print: ({ node, path, print }) => {
     const doc = [
       path.call(print, 'expression'),
-      indent([softline, '.', node.memberName])
+      shallIndent(path)
+        ? indent([softline, '.', node.memberName])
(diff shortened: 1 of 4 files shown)


fvictorio:
This is not quite how prettier formats equivalent statements. This is what prettier does (I only removed the `address` type, otherwise is the same code):

```js
signer = _hashTypedDataV1(
  keccak256(
    abi.encode(
      TYPEHASH,
      req.from,
      req.to,
      req.value,
      req.gas,
      req.nonce,
      keccak256(req.data)
    )
  )
)
  ._hashTypedDataV2(
    keccak256(
      abi.encode(
        TYPEHASH,
        req.from,
        req.to,
        req.value,
        req.gas,
        req.nonce,
        keccak256(req.data)
      )
    )
  )
  ._hashTypedDataV3(
    keccak256(
      abi.encode(
        TYPEHASH,
        req.from,
        req.to,
        req.value,
        req.gas,
        req.nonce,
        keccak256(req.data)
      )
    )
  )
  .recover(signature);

signer = _hashTypedDataV1(
  keccak256(
    abi.encode(
      TYPEHASH,
      req.from,
      req.to,
      req.value,
      req.gas,
      req.nonce,
      keccak256(req.data)
    )
  )
)
  ._hashTypedDataV2(
    keccak256(
      abi.encode(
        TYPEHASH,
        req.from,
        req.to,
        req.value,
        req.gas,
        req.nonce,
        keccak256(req.data)
      )
    )
  )
  ._hashTypedDataV3(
    keccak256(
      abi.encode(
        TYPEHASH,
        req.from,
        req.to,
        req.value,
        req.gas,
        req.nonce,
        keccak256(req.data)
      )
    )
  )
  .recover(signature);
return _nonces[req.from] == req.nonce && signer == req.from;
```

I'm ok with merging this if it's an improvement over what we already have, but I think we should try to have that same output eventually.

Janther:
It's an improvement and still a few to come but it's some work right now. Got some inspiration but am letting it simmer before coding

PR #520, opened 2021-05-30
Assignment in the same line

Janther:
Assignment and variable declarations with an initial value had a slightly different behaviour.
This standardizes this.

There's still room for improvement here but need to start from the same base for those improvements.


mattiaerre (approved):
it makes sense to me thanks @Janther

PR #523, opened 2021-06-01
End of chain

Janther:
A redesign of the `isBeginnigOfChain` method now considering cases like:

```Solidity
a()().b
c[1][2].d
```

While the previous would return false positives since it only cared about
```Solidity
a().b
c.d
```

### DIFF
--- a/src/nodes/MemberAccess.js
+++ b/src/nodes/MemberAccess.js
@@ -4,7 +4,6 @@ const {
   }
 } = require('prettier/standalone');
 
-
 const isEndOfChain = (node, path) => {
   let i = 0;
   let currentNode = node;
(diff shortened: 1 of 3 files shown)


src/nodes/MemberAccess.js
```diff
@@ -4,15 +4,42 @@ const {
...
-    const grandParentNodeType = path.getParentNode(1).type;
-    return grandParentNodeType !== 'MemberAccess';
-  }
+    // If direct ParentNode is a FunctionCall and currentNode is not the expression
+    // then it must be and argument in which case it is the end of the chain.
+    if (
+      parentNode.type === 'FunctionCall' &&
+      currentNode !== parentNode.expression
+    )
+      break;
```
mattiaerre, line 21: can we use return instead of break?

src/nodes/MemberAccess.js
```diff
@@ -4,15 +4,42 @@ const {
...
 
+    if (
+      parentNode.type === 'BinaryOperation' ||
+      parentNode.type === 'UnaryOperation' ||
+      parentNode.type === 'IfStatement' ||
+      parentNode.type === 'WhileStatement' ||
+      parentNode.type === 'ForStatement' ||
+      parentNode.type === 'VariableDeclarationStatement' ||
+      parentNode.type === 'ExpressionStatement'
```
mattiaerre, line 35: can we alpha order these?

mattiaerre (commented):
do you think we should add additional test/use case for this PR?

Janther:
Yes if it comes to mind please do.
Also if you could find a way for using Prettier's ASTPath in testing without going through the whole process of parsing, that would be much better.

src/nodes/MemberAccess.js
Janther: If we return, we might get to a point where we add all of the possible scenarios and the last return never runs (but we need to keep it for safety) thus breaking our 100% streak.

src/nodes/MemberAccess.js
fvictorio: This list of nodes feel a bit ad hoc, what's the reason behind them?

Btw, one way to both simplify this code _and_ document its intention a bit better is to do something like this:

```js
const nodesWhereWeBreakForSomeReason = ['BinaryOperation', 'UnaryOperation', ...]
if (nodesWhereWeBreakForSomeReason.includes(parentNode.type)) {
  break
}
```

where `nodesWhereWeBreakForSomeReason` has a proper name :smile:

fvictorio (approved):

PR #524, opened 2021-06-01
Function calls set group

Janther:
`FunctionCall`s now tell the chain if it breaks so we can choose whether to have a new line or not.


mattiaerre (approved):
it LGTM; qq do you think we should have a `constants` file where we keep all the magi strings that we have in the other files?

PR #531, opened 2021-06-07
By flattening the array we have a better analysis of the output

Janther:
Also ditching the `label` since we used the data we needed.


Janther:
This change is internal and for us to have some piece of mind

mattiaerre (approved):
👍

PR #532, opened 2021-06-07
using the new test organisation from prettier

Janther:
the organisation makes a lot of sense and gives us a standard.


Janther:
if anyone had an idea of how to move the contents of `tests_config` to `tests/config` that would be great as I had no success.

.prettierignore
```diff
@@ -1,4 +1,3 @@
 coverage/**/*.js
-tests/**/*.sol
-tests_config/**/*.js
```
mattiaerre, line 3: should we add `tests/config` to this file too?

tests/config/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "singleQuote": false,
+    "trailingComma": "es5"
+  }
+
```
mattiaerre, line 5: can you add a new line here? thanks

tests/config/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+    "singleQuote": false,
```
mattiaerre, line 2: double quotes now? really?

tests/config/format-test.js
```diff
@@ -1,30 +1,32 @@
+"use strict";
```
mattiaerre, line 1: why did we add this? is this not always the case?

tests/config/.prettierrc
```diff
@@ -0,0 +1,5 @@
...
+    "trailingComma": "es5"
```
mattiaerre, line 3: and trailing comma too? oh noes 😱

tests/config/require-prettier.js
```diff
@@ -0,0 +1,5 @@
+"use strict";
+
+const prettier = require("prettier");
+
+module.exports = prettier;
```
mattiaerre, line 5: what is this file for?

tests/unit/binary-operator-printers/index.test.js
```diff
@@ -1,4 +1,4 @@
-const binaryOperatorPrinters = require('../../src/binary-operator-printers');
+const binaryOperatorPrinters = require('../../../src/binary-operator-printers');
```
mattiaerre, line 1: single quotes here?

mattiaerre (commented):
added some comments, I will add more later I think

mattiaerre:
@Janther can you add a link to where this "inspiration" is coming from? I think I am missing some context here 😊

.prettierignore
Janther: On this folder I added the `.prettierrc` with the default that prettier uses so we still benefit from using prettier on the small changes that we make here for us while keeping their code untouched.

Janther: eslint on the other hand should ignore this folder

tests/config/.prettierrc
Janther: this is just for the files we took from the prettier repo using prettier's defaults

tests/config/format-test.js
Janther: I want to keep these files as close as possible to prettier's so it's easier to update them following their new practices

tests/config/.prettierrc
Janther: same as the other comment

tests/unit/binary-operator-printers/index.test.js
Janther: this is our tests with our `.prettierrc` the default config is only for `tests/config`

tests/config/require-prettier.js
Janther: the tests can run using `prettier` or `prettier/standalone` as the engine.

Janther:
> @Janther can you add a link to where this "inspiration" is coming from? I think I am missing some context here 😊

there are a lot of things that we decide on how to do regarding the formatting of code. But we lack some vision in 2 aspects:

  * Comments
  * Testing

That is why we are using `prettier`s code for these and as long as we do this, I'd like to keep a close look at what they are doing in these aspects and not be stale.

fvictorio (approved):

PR #535, opened 2021-06-11
Compiler option

Janther:
We added the option to specify the compiler we want prettier to use.

we can specify a particular version and we'll log a warning if the `PragmaDirective`s don't support the chosen compiler.

Now that we have knowledge of the compiler being used, we can make some formatting choices.

We can decide whether we should split an `ImportDirective`.

- If no compiler is given we will not split the `ImportDirective`.
- If the given compiler satisfies `<0.7.4` we will not split the `ImportDirective`.
- If the given compiler satisfies `>=0.7.4` we will not split the `ImportDirective` if necessary (as the bug that didn't recognise a split import was fixed here).

Also the type `byte` was used by `explicitTypes` comes back.

- If no compiler is given we enforce `bytes1`.
- If the given compiler satisfies `<0.8.0` we rely on the `explicitTypes` option.
- If the given compiler satisfies `>=0.8.0` we enforce `bytes1`.

Finally we are supporting adding parentheses to exponentiation of exponentiation operations.
- If no compiler is given we keep `a**b**c`.
- If the given compiler satisfies `<0.8.0` we format `(a**b)**c`.
- If the given compiler satisfies `>=0.8.0` we format `a**(b**c)`.


src/common/util.js
```diff
@@ -0,0 +1,11 @@
+const getCompiler = (path) => {
+  let i = 0;
+  let elderNode;
+  do {
+    elderNode = path.getParentNode(i);
+    i += 1;
+  } while (elderNode.type !== 'SourceUnit');
```
fvictorio, line 7: This assumes that `SourceUnit` will always be the root node, which I think is a fair assumption. But I'd rather codify this assumption:

```
} while (elderNode.type && elderNode.type !== 'SourceUnit')

if (!elderNode || !elderNode.compiler) {
  throw new Error("AssertionError: expected elderNode.compiler to exist")
}

Comment (Janther) [reply src/common/util.js]: we don't use this anymore

File: .gitignore [hunk 2]
```diff
@@ -62,6 +62,10 @@ typings/
...
 
-# package-lock.json
-
+# yarn 2 config
 yarn.lock
+.yarn
+.yarnrc.yml
+
+# vscode config
+.vscode
```
mattiaerre, line 71: add new line please 😊

src/nodes/print-preserving-empty-lines.js
```diff
@@ -12,9 +12,9 @@ function printPreservingEmptyLines(path, key, options, print) {
     const nodeType = node.type;
 
     if (
-      // Avoid adding a hardline at the beggining of the document.
+      // Avoid adding a hardline at the beginning of the document.
       parts.length !== 0 &&
-      // LabelDefinition adds a dedented line so we have to avoid prepending
+      // LabelDefinition adds a dedented line so we have don't have to prepend
```
mattiaerre, line 17: remove the 1st `have`

src/common/util.js
```diff
@@ -0,0 +1,6 @@
+const getCompiler = (options) => {
+  if (!options.compiler) return undefined;
+  return options.compiler.split(/-|\+/)[0];
+};
```
mattiaerre, line 4: are we documenting this somewhere?

src/nodes/FunctionDefinition.js
```diff
@@ -40,7 +40,7 @@ const parameters = (parametersType, node, path, print, options) => {
   }
   if (node.comments && node.comments.length > 0) {
     // we add a check to see if the comment is inside the parentheses
-    const paremeterComments = printComments(
+    const parameterComments = printComments(
```
mattiaerre, line 43: 👍

src/common/util.js
```diff
@@ -0,0 +1,6 @@
...
+
+module.exports = { getCompiler };
```
mattiaerre, line 6: do you think there will be more utils? 😅

src/nodes/PragmaDirective.js
```diff
@@ -1,18 +1,5 @@
...
-      .replace(/([<>=])/g, ' $1')
-      .replace(/< =/g, '<=')
-      .replace(/> =/g, '>=')
-      .trim();
-    if (value.split(' ').length > 1) {
-      value = semver.validRange(value);
-    }
-    return ['pragma ', node.name, ' ', value, ';'];
-  }
+  print: ({ node }) => ['pragma ', node.name, ' ', node.value, ';']
```
mattiaerre, line 2: much nicer

mattiaerre (commented):
I think we should add in the readme a section where we explain the options the plugin takes if we haven't yet

src/common/util.js
Janther: I'm planning on moving some files from the `nodes` folder to `common` and there are some utils in the `prettier-comments/utils` that I'm preparing to clean up and move here. Since the long term goal is to delete this very outdated "dependency".

src/common/util.js
Janther: Done

src/nodes/print-preserving-empty-lines.js
Janther: done

.gitignore
Janther: done

Janther:
> I think we should add in the readme a section where we explain the options the plugin takes if we haven't yet

Yup I definitely agree with this one

src/parser.js
```diff
@@ -12,11 +13,36 @@ const tryHug = (node, operators) => {
...
+      if (ctx.name !== 'solidity') return;
+      // if the compiler option has not been provided we leave.
+      if (!compiler) return;
+      // we make a check against each pragma directive in the document.
+      if (!semver.satisfies(compiler, ctx.value)) {
+        // @TODO: investigate the best way to warn that would apply to
+        // different editors.
+        // eslint-disable-next-line no-console
+        console.warn(
+          `The compiler '${options.compiler}' does not satisfy 'pragma solidity ${ctx.value}'.`
```
fvictorio, line 42:
```suggestion
          `[prettier-solidity] The compiler option is set to '${options.compiler}', which does not satisfy 'pragma solidity ${ctx.value}'.`
```

fvictorio:
> If no compiler is given we will not split the ImportDirective.

> If no compiler is given we enforce bytes1.

@Janther that seems inconsistent to me. In the first case, we are doing the conservative thing and doing what we would do for an "old" compiler version. In the second case though, we are doing something that only works for solidity >= 0.8.0. Am I missing something here?

Janther:
`bytes1` works on old and new.
`byte` works only on old.
The strategy here is: if we don't know we go with the safest. As a matter of facts all 3 formats we do use different approaches when we don't know the compiler:
- Imports: we choose to keep everything in one line (old compiler strategy)
- Exponentiation: we choose not to add parentheses (do nothing strategy)
- `bytes1`: we choose to enforce `bytes1` since `byte` only works in old compilers (new compiler strategy)

We could leave `byte` and `bytes` alone and have the error be handled by the linter or the compiler and choose a **do nothing strategy**. But enforcing the safest format seems better to me.

fvictorio:
> `bytes1` works on old and new.
> `byte` works only on old.
> The strategy here is: if we don't know the compiler we go with the safest format so it works on both.

Ugh, you are correct, sorry, I misremembered that. This makes sense to me!

fvictorio:
I think the only remaining question I have is... should we mark this option as experimental? I'm inclined to think that we should since this is a significant change and we haven't battle-tested this approach.

Janther:
Feel free to suggest those changes

PR #537, opened 2021-06-15
vim-plug plugin dir changed

TimDaub:
vim-plug now uses `plugged` as plugin directory name

- May help to address problem in #537


fvictorio:
Thanks @TimDaub. I also removed the `branch` part because I don't think that's valid anymore.

fvictorio (approved):

PR #539, opened 2021-06-16
adding missing metadata for the language.

Janther:
Info taken from [linguist-languages](https://github.com/ikatyang/linguist-languages/blob/master/data/Solidity.json).

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,9 +9,14 @@ const print = require('./printer');
 // https://prettier.io/docs/en/plugins.html#languages
 const languages = [
   {
+    aceMode: 'text',
+    color: '#AA6746',
     extensions: ['.sol'],
+    linguistLanguageId: 237469032,
     name: 'Solidity',
     parsers: ['solidity-parse'],
+    tmScope: 'source.solidity',
+    type: 'programming',
     vscodeLanguageIds: ['solidity']
   }
 ];
--- a/src/index.js
+++ b/src/index.js
@@ -6,6 +6,7 @@ const parse = require('./parser');
 const print = require('./printer');
 
 // https://prettier.io/docs/en/plugins.html#languages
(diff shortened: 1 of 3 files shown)


Janther:
also we could have `linguist-languages` as a dependency and merge their json with a much leaner amount of data in our code.
```Javascript
solidityLang = require("linguist-languages/data/Solidity.json");

language = {
  ...solidityLang,
  since: "0.0.0",
  parsers: ['solidity-parse'],
  vscodeLanguageIds: ['solidity']
}
```

src/index.js
```diff
@@ -1,16 +1,18 @@
...
 const print = require('./printer');
 
 // https://prettier.io/docs/en/plugins.html#languages
+const { languageId, ...rest } = linguistData;
 const languages = [
   {
-    extensions: ['.sol'],
-    name: 'Solidity',
+    linguistLanguageId: languageId,
+    ...rest,
```
mattiaerre, line 14: I would like to keep it explicit or else add what's in the `linguistData` object if possible

src/index.js
```diff
@@ -1,16 +1,18 @@
...
+    since: '0.0.0',
```
mattiaerre, line 15: is this required?

src/index.js
Janther: can do the object is very small.

src/index.js
Janther: it's optional but am unsure if it's prettier's version, or our plugin's version

mattiaerre (approved):
thanks @Janther one very minor improvement would be to alpha order the properties but I'm happy to merge as is

Janther:
> thanks @Janther one very minor improvement would be to alpha order the properties but I'm happy to merge as is
> 
> 

I'm keeping the properties like I got them from the linguist object for the moment we need to update this. Then we only need to copy paste the data on top.

PR #540, opened 2021-06-17
Using prettier instead of prettier/standalone and let webpack decide which to use

Janther:
From Prettier's [documentation](https://prettier.io/docs/en/browser.html).

> The browser field in Prettier’s package.json points to standalone.js. That’s why you can just import or require the prettier module to access Prettier’s API, and your code can stay compatible with both Node and the browser as long as webpack or another bundler that supports the browser field is used. This is especially convenient for plugins.

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -1,4 +1,5 @@
 const FULL_TEST = Boolean(process.env.FULL_TEST);
+const TEST_STANDALONE = Boolean(process.env.TEST_STANDALONE);
 
 module.exports = {
   collectCoverage: FULL_TEST,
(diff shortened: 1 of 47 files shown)


Janther:
From Prettier's documentation.

> The browser field in Prettier’s package.json points to standalone.js. That’s why you can just import or require the prettier module to access Prettier’s API, and your code can stay compatible with both Node and the browser as long as webpack or another bundler that supports the browser field is used. This is especially convenient for plugins.

jest.config.js
```diff
@@ -1,4 +1,5 @@
 const FULL_TEST = Boolean(process.env.FULL_TEST);
+const TEST_STANDALONE = Boolean(process.env.TEST_STANDALONE);
```
mattiaerre, line 2: where this env variable will be used? I cannot see it in this PR

src/nodes/print-separated-list.js
```diff
@@ -14,7 +14,7 @@ const {
 const printSeparatedList = (
   list,
   {
-    groupId = undefined,
```
mattiaerre, line 17: 👍

jest.config.js
Janther: this variable is used to run tests.
We don't actively use in our automated tests right now but the idea is to have an environment that uses only a browser friendly environment.

PR #549, opened 2021-07-01
Respecting prettiers default options

Janther:
For correctness I'm adding back all of the options supported by our plugin, so other tools can use this information (e.g. generate documentation or generate a form for all of the options)

These where removed in the past because they were overriding Prettier's default options for other plugins (see #375).
I added a test to see that this isn't the case this time. It works well but Standalone mode doesn't have a strong default strategy (I reported an issue prettier/prettier#11107).


fvictorio (approved):
LGTM. I added an (unrelated but minor) change to the description of the `explicitTypes` options.

PR #552, opened 2021-07-09
Throwing if the prettier version is not >=2.3.0

Janther:
Preventing from printing with an old version of prettier that still relies on `concat`.

### DIFF
--- a/tests/unit/prettier-version.test.js
+++ b/tests/unit/prettier-version.test.js
@@ -12,7 +12,7 @@ jest.mock('prettier', () => {
 
 const prettier = require('prettier');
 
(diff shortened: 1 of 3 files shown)


tests/unit/prettier-version.test.js
```diff
@@ -0,0 +1,26 @@
...
+
+  return {
+    ...originalModule,
+    version: '2.2.1'
+  };
+});
+
+const prettier = require('prettier');
+
+test('should throw if the installed version of prettier is lesser than v2.3.0', () => {
```
mattiaerre, line 15: s/lesser/less

mattiaerre (approved):
this is fantastic, thanks @Janther

Issue #554, opened 2021-07-13
Parentheses on modifiers without arguments


frangio:
I expected prettier-solidty to have an opinion about parentheses on modifiers without arguments. To my knowledge, they are generally used _without_ parentheses and so they should be omitted.

```solidity
function foo() onlyOwner public {
// versus
function foo() onlyOwner() public {
```

This is related to https://github.com/prettier-solidity/prettier-plugin-solidity/issues/165 about modifier definitions, which was closed but no decision seems to have been made although it was part of the issue:
> Since prettier is supposed to be an opinionated formatter I propose that [...] prettier-solidity should always print modifier declarations without parentheses.

I agree that prettier-solidity should be opinionated about this, both for modifier definitions and uses.

Janther:
Addressed in our latest [release](https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-beta.16)

Issue #165, opened 2019-06-23
Modifier Declarations without parameters should not have parentheses


Janther:
`solidity-parser-antlr`'s new [version](https://github.com/federicobond/solidity-parser-antlr/pull/77) will introduce a breaking change by removing `Parameter` and `ParameterList` and using a collection of `VariableDeclaration` instead.
I made the changes and discovered that we no longer will be able to know when a modifier without parameters has been declared like this:
```Solidity
modifier a {
    _;
}
```
or like this:
```Solidity
modifier a() {
    _;
}
```
for the time being, I asked @federicobond to add the case that if no parentheses where given, the parser could show `null` instead of an empty array `[]`.
He agreed but is a bit reluctant about this change.
This solution was inspired by the use case where constructor modifiers need to preserve parentheses otherwise the code will not compile.

In the constructor modifier, this is necessary until @federicobond finds a better way to describe that a modifier is a constructor call.
In a modifier declaration this solution is not really needed as both declarations (with and without parentheses) generate the same bytecode.

Since prettier is supposed to be an opinionated formatter I propose that the parser should not take care of this use case and prettier-solidity should always print modifier declarations without parentheses.

That way the parser will be closer to an AST rather a Concrete Syntax Tree (which was @federicobond's issue with this change). And we would still offer a propper prettified code.

fvictorio:
This seems to work fine now.

PR #555, opened 2021-07-13
removing parentheses from modifiers with an empty list of parameters.

Janther:
addresses #554 and #165

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -44,8 +44,8 @@ function parse(text, _parsers, options) {
       }
     },
     ModifierDefinition(ctx) {
-      if (ctx.parameters && ctx.parameters.length === 0) {
-        ctx.parameters = null;
+      if (!ctx.parameters) {
+        ctx.parameters = [];
       }
     },
     FunctionDefinition(ctx) {
--- a/src/parser.js
+++ b/src/parser.js
@@ -48,9 +48,14 @@ function parse(text, _parsers, options) {
         ctx.parameters = null;
       }
     },
-    ModifierInvocation(ctx) {
-      if (ctx.arguments && ctx.arguments.length === 0) {
-        ctx.arguments = null;
+    FunctionDefinition(ctx) {
+      if (!ctx.isConstructor) {
+        ctx.modifiers.forEach((modifier) => {
+          if (modifier.arguments && modifier.arguments.length === 0) {
+            // eslint-disable-next-line no-param-reassign
+            modifier.arguments = null;
+          }
+        });
       }
     },
     ForStatement(ctx) {
--- a/src/parser.js
+++ b/src/parser.js
@@ -43,6 +43,16 @@ function parse(text, _parsers, options) {
         );
       }
     },
+    ModifierDefinition(ctx) {
+      if (ctx.parameters && ctx.parameters.length === 0) {
+        ctx.parameters = null;
+      }
+    },
+    ModifierInvocation(ctx) {
+      if (ctx.arguments && ctx.arguments.length === 0) {
+        ctx.arguments = null;
+      }
+    },
     ForStatement(ctx) {
       if (ctx.initExpression) {
         ctx.initExpression.omitSemicolon = true;
--- a/tests/format/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
@@ -1033,23 +1033,23 @@ contract Overrides {
             BazQuxBazQuxBazQux,
             Quuuuuuuuuuuuuux
         ) varName;
-    modifier onlyOwner() virtual {
+    modifier onlyOwner virtual {
         _;
     }
 
-    modifier onlyOwner() override {
+    modifier onlyOwner override {
         require(msg.sender == owner(), "Ownable: caller is not the owner");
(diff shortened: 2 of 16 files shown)


tests/format/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -1033,23 +1033,23 @@ contract Overrides {
             BazQuxBazQuxBazQux,
             Quuuuuuuuuuuuuux
         ) varName;
-    modifier onlyOwner() virtual {
+    modifier onlyOwner virtual {
```
fvictorio, line 1036: I had no idea you could omit the parentheses in the modifier _definition_. I think we should keep them in the definitions but erase them in the invocations. @frangio what do you think?

frangio: That would be aligned with what we do in OpenZeppelin Contracts. Personally I'm ok with that!

Janther: mmmh having a look at this I believe we can force the definitions to always have a parentheses but for invocations we must leave these alone since we can't differentiate between a modifier invocation and a constructor invocation i.e. `Ownable()`

frangio: Is it possible to at least remove parentheses when the function is clearly not a constructor? With the consideration that older versions of Solidity used `function ContractName` syntax rather than `constructor`, though I don't know if those are supported by prettier-solidity.

Janther: Yeah yeah. Been dwelling on the same. We can do that.

Janther: Ok so this is the new behaviour:
  - ModifierDefinitions with no parameters will always have an empty parentheses.
  - ModifierInvocations with no arguments belonging to a constructor will keep the same parentheses syntax from the input.
  - ModifierInvocations with no arguments not belonging to a constructor will never print parentheses.
I also added a test to compare the `bytecode` to ensure that no harm is done by these changes.

fvictorio: Agh, that sucks, but there's no way to avoid that because the parser doesn't have type information. I think that trade-off is ok, but let's not merge this yet because I want to wait for @frangio's input and because I want to add an "edge cases" section to the readme to document this (I'll do that tomorrow).

src/parser.js
```diff
@@ -43,6 +43,21 @@ function parse(text, _parsers, options) {
         );
       }
     },
+    ModifierDefinition(ctx) {
+      if (!ctx.parameters) {
+        ctx.parameters = [];
```
Janther, line 48: Why doesn't `eslint` warn about `no-param-reassign` here like in line 56?

Janther:
@fvictorio, now that we have an opinion around modifier definitions, maybe the parser could only return an array  for `ModifierDefinition`'s `parameters` attribute and have a more restricted type.

Quoting @federicobond https://github.com/federicobond/solidity-parser-antlr/pull/77#issuecomment-504275414

src/parser.js
mattiaerre: not sure, maybe because of this? https://eslint.org/docs/rules/no-param-reassign#options

mattiaerre (approved):
👍

src/parser.js
Janther: it's an exception of `airbnb-base`
https://github.com/airbnb/javascript/blob/f5c14cae2ff58000cead98290b8ec4b54dda2f14/packages/eslint-config-airbnb-base/rules/best-practices.js#L237

tests/format/FunctionDefinitions/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -17,11 +17,11 @@ interface FunctionInterfaces {
 
   function manyParams(uint x1, uint x2, uint x3, uint x4, uint x5, uint x6, uint x7, uint x8, uint x9, uint x10);
 
-  function manyModifiers() modifier1 modifier2 modifier3 modifier4 modifier5 modifier6 modifier7 modifier8 modifier9 modifier10;
+  function manyModifiers() modifier1() modifier2() modifier3 modifier4 modifier5 modifier6 modifier7 modifier8 modifier9 modifier10;
```
frangio, line 20: Shouldn't these parentheses be removed?

tests/format/AllSolidityFeatures/__snapshots__/jsfmt.spec.js.snap
frangio: I think this behavior is good.

tests/format/FunctionDefinitions/__snapshots__/jsfmt.spec.js.snap
fvictorio: This is the "input" part of the snapshot. The output, that didn't change (and therefore it's not shown here) doesn't have parentheses in these modifiers.

fvictorio (approved):

Issue #556, opened 2021-07-14
indentation of method call


zhiji666:
now it's formatted as
````
int256 amount = SafeCast
.toInt256(amount.mul(10**(18 - underlyingAssetDecimals)))
.neg();
````
maybe 4 spaces is better
````
int256 amount = SafeCast
    .toInt256(amount.mul(10**(18 - underlyingAssetDecimals)))
    .neg();
````

Janther:
Addressed in our latest [release](https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-beta.16)

PR #557, opened 2021-07-15
Chained member access

Janther:
With this PR I'm rethinking the way of how we print chains. This PR has been in my roadmap for a while and was only possible after #523 and #531.

Here we add another function to process the doc.

`processChain` expects the flattened `doc[]` of the complete chain of MemberAccess.

It uses the separator label to split the chain into 2 arrays.
The first array is the `doc[]` corresponding to the first element before the first separator.
The second array contains the rest of the chain.

The indentation of the whole chain depends on the result of the first element.

If the first element breaks into multiple lines, we won't indent the rest of the chain as the last line (most likely a closing parentheses) won't be indented.

i.e.
```Solidity
functionCall(
    arg1,
    arg2
).rest.of.chain

functionCall(
    arg1,
    arg2
)
.long
.rest
.of
.chain
```

If the first element doesn't break into multiple lines we treat the rest of the chain as a normal chain and proceed to indent it.

i.e.
```Solidity
functionCall(arg1, arg2).rest.of.chain

functionCall(arg1, arg2)
    .long
    .rest
    .of
    .chain
```

NOTE: As described in the examples above, the rest of the chain will be grouped and try to stay in the same line as the end of the first element.

Chained `MemberAccess` where always clunky since all of the steps had their own indentation and we where juggling to make things look nice.

I believe that these changes will make the work with chains easier.

PS: Also this PR contains the possibility to have an `IndexAccess` as the beginning of the chain.

```Solidity
veryLongNamedArray[
    aStructVariable.indexValue
].rest.of.chain

veryLongNamedArray[
    aStructVariable.indexValue
]
.long
.rest
.of
.chain
```

### DIFF
--- a/tests/format/FunctionCalls/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/FunctionCalls/__snapshots__/jsfmt.spec.js.snap
@@ -52,7 +52,7 @@ contract FunctionCalls {
 contract FunctionCalls {
     function foo() {
         address veryLongValidatorAddress = veryVeryVeryLongSignature
-        .popLast20Bytes();
+            .popLast20Bytes();
     }
 
     function foo() {
@@ -80,7 +80,8 @@ contract FunctionCalls {
                     keccak256(req.data)
                 )
             )
-        )._hashTypedDataV2(
+        )
+        ._hashTypedDataV2(
             keccak256(
                 abi.encode(
                     TYPEHASH,
@@ -92,7 +93,8 @@ contract FunctionCalls {
                     keccak256(req.data)
                 )
             )
-        )._hashTypedDataV3(
+        )
+        ._hashTypedDataV3(
             keccak256(
                 abi.encode(
                     TYPEHASH,
@@ -104,7 +106,8 @@ contract FunctionCalls {
                     keccak256(req.data)
                 )
             )
-        ).recover(signature);
+        )
+        .recover(signature);
         signer = _hashTypedDataV1(
             keccak256(
                 abi.encode(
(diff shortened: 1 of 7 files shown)


Janther:
By the way this PR completely overkills #556 in a much needed way.

passabilities:
yes please I'm losing my mind with the indents

tests/format/FunctionCalls/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -80,7 +80,8 @@ contract FunctionCalls {
                     keccak256(req.data)
                 )
             )
-        )._hashTypedDataV2(
+        )
+        ._hashTypedDataV2(
```
fvictorio, line 84: This is definitely an improvement, but FWIW prettier core would've indented this and the rest of the chain (which makes sense to me).

No need to change this in this PR, we can open a new issue for that.

fvictorio (approved):
Fantastic improvement. The code went over my head, to be honest, but I get the basic idea.

tests/format/FunctionCalls/__snapshots__/jsfmt.spec.js.snap
Janther: Are you sure? I did cross check with prettier when I was designing the logic behind this.
I'll merge and we can give it another review later

Janther: Wow I had another check and indeed it does indent. This really simplifies a lot for this behaviour then 🤣 Let's open an issue for this though.

Issue #559, opened 2021-07-15
Indent long chains


fvictorio:
This is the current format we are printing:

```solidity
  uint foo = someFunction(
    foo,
    bar,
    baz
  )
  .someOtherFunction(
    foo,
    bar,
    baz
  )
```

while prettier core does something more similar to this:

```solidity
  uint foo = someFunction(
    foo,
    bar,
    baz
  )
    .someOtherFunction(
      foo,
      bar,
      baz
    )
```

See this [discussion](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/557#discussion_r670732802) for more context.

Janther:
Also I'd like to say that currently we don't indent only when the first object of the chain splits into multiple lines. If it stays in a single line we format like this:

```Solidity
  uint foo = someFunction(foo, bar, baz)
    .someOtherFunction(
      foo,
      bar,
      baz
    )
```

I like the continuity of the chain we get when the first object splits and we end with a single line having a closing `)` or `]` and right bellow it we see the next item.

This approach although nice, adds some coupling between different nodes in a sense that a `FunctionCall` has to tell the `MemberChain` some extra info about itself.

The formatting used by Prettier allows us to be decoupled.

I personally am happy with both and already provided a PR #560 to move to the new format if that's what we want.

mattiaerre:
I think we can close this one given that #560 has been merged, would you agree @fvictorio and @Janther ?

Janther:
There are things that I'd like to leave open for the community to discuss.
Maybe an issue is not great and a better way would be a different platform.
@fvictorio??

fvictorio:
I tend to err on the side of closing issues instead of keeping them open. If someone has something to say, they can open a new issue, let us know in the telegram group, or just comment here since we read comments on closed issues too.

That being said, feel free to re-open if you disagree!

PR #560, opened 2021-07-15
Chain behaviour closer to prettier core

Janther:
with this new behaviour there's a lot of decoupling between the nodes and allows us to clean the code.
We don't need `label`s to pass on metadata across different nodes (although we still use them)

### DIFF
--- a/src/nodes/MemberAccess.js
+++ b/src/nodes/MemberAccess.js
@@ -1,6 +1,6 @@
 const {
   doc: {
-    builders: { group, ifBreak, indent, label, softline }
+    builders: { group, indent, label, softline }
   }
 } = require('prettier');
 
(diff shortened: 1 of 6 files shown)


Janther:
addresses #559

fvictorio (approved):

mattiaerre:
> addresses #559

if you you `closes` as a keyword then once you merge the PR the issue is automatically closed

src/nodes/MemberAccess.js
```diff
@@ -55,15 +55,22 @@ const isEndOfChain = (node, path) => {
  * first separator.
  * The second array contains the rest of the chain.
  *
- * The indentation of the whole chain depends on the result of the first
- * element.
+ * The second array is grouped and indented and while the first element's
```
mattiaerre, line 58: you may want to remove `and`

mattiaerre (approved):
👍

PR #561, opened 2021-07-16
Giving a more detailed step by step description on using VSCode

Janther:
There is still some confusion on how to use the latest version of our plugin with VSCode so I added more details

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,21 @@ code --install-extension JuanBlanco.solidity
 
 This extension provides basic integration with Prettier for most cases no further action is needed.
 
+Make sure your editor has format on save set to true.
+When you save VSCode will ask you what formatter would you like to use for the solidity language, you can choose `JuanBlanco.solidity`.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -270,10 +285,28 @@ To interact with 3rd party plugins, `prettier-vscode` will look in the project's
 npm install --save-dev prettier prettier-plugin-solidity
 ```
 
-This will allow you to specify the version of the plugin in case you need to freeze the formatting since new versions of this plugin will implement tweaks on the possible formats.
+This will allow you to specify the version of the plugin in case you want to use the latest version of the plugin or need to freeze the formatting since new versions of this plugin will implement tweaks on the possible formats.
 
 You'll have to let VSCode what formatter you prefer.
-As a final check, make sure that VSCode is configured to format files on save.
+This can be done by opening the command pallette and executing:
```
mattiaerre, line 291: s/pallette/palette

mattiaerre (approved):
one minor comment

PR #565, opened 2021-07-20
solving the indentation at the beginning of a var declaration that splits

### LINKED ISSUE #564: Weird formatting behavior on line with a struct var declaration
Before format: 

```
    function isAuthorized(
        bytes32 serviceId,
        address client
    ) external view override returns (bool) {
        WhitelistStatus storage whitelistStatus = serviceIdToClientToWhitelistStatus[serviceId][client];
        return true;
    }
```

After format: 

```
  function isAuthorized(bytes32 serviceId, address client)
    external
    view
    override
    returns (bool)
  {

      WhitelistStatus storage whitelistStatus
     = serviceIdToClientToWhitelistStatus[serviceId][client];
    return true;
  }
```

Not sure why but a **new line** and an **extra tab** is added to the first line in the function.

In .prettierrc I have the following overrides set:

```
      {
        "files": "*.sol",
        "options": {
          "compiler": "0.8.6",
          "printWidth": 80,
          "tabWidth": 4,
          "useTabs": false,
          "singleQuote": false,
          "bracketSpacing": false,
          "explicitTypes": "always"
        }
      }
```

Janther:
closes #564 

This revealed 2 issues

  - The variables where printed as a indented list even if it was only one.
  - If the declaration splits, the indentation of the initial value was not updated.

The Index access looks weird though, maybe we have to create an issue for this.

### DIFF
--- a/src/nodes/VariableDeclarationStatement.js
+++ b/src/nodes/VariableDeclarationStatement.js
@@ -1,6 +1,6 @@
 const {
   doc: {
-    builders: { group }
+    builders: { group, ifBreak, indent }
   }
 } = require('prettier');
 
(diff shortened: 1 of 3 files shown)


fvictorio:
For future reference, this is what prettier core does:

```ts
// print-width: 100
function isAuthorized() {
  whitelistStatus: WhitelistStatus = serviceIdToClientToWhitelistStatus[serviceId][client];
  return true;
}

// print-width: 80
function isAuthorized() {
  whitelistStatus: WhitelistStatus =
    serviceIdToClientToWhitelistStatus[serviceId][client];
  return true;
}

// print-width: 50
function isAuthorized() {
  whitelistStatus: WhitelistStatus =
    serviceIdToClientToWhitelistStatus[serviceId][
      client
    ];
  return true;
}

// print-width: 40
function isAuthorized() {
  whitelistStatus: WhitelistStatus =
    serviceIdToClientToWhitelistStatus[
      serviceId
    ][client];
  return true;
}

// print-width: 10
function isAuthorized() {
  whitelistStatus: WhitelistStatus =
    serviceIdToClientToWhitelistStatus[
      serviceId
    ][
      client
    ];
  return true;
}
```

I'm ok with moving it to an issue though.

fvictorio (approved):

Janther:
I'm ok without having this in an issue for now. It complicates the rules a lot to try to mimic Prettier in this case and the benefits aren't that great.

PR #570, opened 2021-08-18
using latest solc

Janther:


package.json
```diff
@@ -76,7 +76,7 @@
     "lines-and-columns": "^1.1.6",
     "outdent": "^0.8.0",
     "prettier": "^2.3.2",
-    "solc": "^0.8.6"
+    "solc": "^0.8.7-fixed"
```
fvictorio, line 79: What does `-fixed` means here?

Janther: It's the official name of the release in npm https://www.npmjs.com/package/solc

They might have done a bad release and then published a new one

fvictorio: I wonder if we should drop the `^` in that case. Not a big deal though.

fvictorio (approved):

PR #643, opened 2022-02-10
Fix "using for" with qualified library name

### LINKED ISSUE #622: Libraries under namespaces are not accepted in `using ... for`
If a library was imported by a qualified import, it cannot be used in `using ... for`.

How to reproduce:

```solidity
// L.sol
library Lib {}
```

```solidity
// C.sol
import "./L.sol" as L;

contract C {
  using L.Lib for uint;
}
```

Running `prettier C.sol` results in:
```
C.sol[error] C.sol: Error: The specified node does not exist
```

Note that with an unqualified import:
```solidity
// C2.sol
import "./L.sol" /* as L */;

contract C2 {
  using /* L. */Lib for uint;
}
```
`prettier C2.sol` runs successfully.

fvictorio:
Fixes #622.

Notice that this changes the version of the `package-lock.json`, which causes a lot of modifications there. I recommend you upgrade `npm` to the latest version so that this isn't changed again.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1684,9 +1684,9 @@
       }
     },
     "@solidity-parser/parser": {
-      "version": "0.14.0",
-      "resolved": "https://registry.npmjs.org/@solidity-parser/parser/-/parser-0.14.0.tgz",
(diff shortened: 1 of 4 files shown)


fvictorio:
Done.

mattiaerre (approved):
👍

PR #685, opened 2022-05-28
memory safe for inline assembly

Janther:


Janther:
@fvictorio Sorry for the delay. Just did.

fvictorio (approved):

Issue #644, opened 2022-02-10
Issue with comment inside property chain


fvictorio:
As reported by @wighawag:

This

```solidity
game.
// CONFIG
config.resolveWindow = 1000;
config.defaultDelay = 0;
```

gets re-written as:

```solidity
game.config.resolveWindow = 1000;
// CONFIG
config.defaultDelay = 0;
```

This seems like a PITA to fix, and I'm not even sure what the "correct" result should be.

Janther:
I believe this should be formatted as

```Solidity
// CONFIG
game.config.resolveWindow = 1000;
config.defaultDelay = 0;
```

Will give this a look.

fvictorio:
Agree, that's that prettier does for the same code in javascript.

fvictorio:
@Janther what should we do about this issue? Is it still valid?

Janther:
Prepending the comments to the beginning of the chain messes up the order of the comments since the chains are printed in reverse order..

```Solidity
game.
// CONFIG
config.
// Window
resolveWindow = 1000;
config.defaultDelay = 0;

// Window
// CONFIG
game.config.resolveWindow = 1000;
config.defaultDelay = 0;
```

We can force the chain to break if any member have comments but the comment will keep the indentation of the member it trails.

```Solidity
game.
// CONFIG
config.
// Window
resolveWindow = 1000;
config.defaultDelay = 0;

game.
// CONFIG
    config.
    // Window
    resolveWindow = 1000;
config.defaultDelay = 0;
```

Also consider that there can be more that one comment per member.

```Solidity
game.
// CONFIG
// Do not touch
config. // Window
// Resolution 1000
resolveWindow = 1000;
config.defaultDelay = 0;

game.
// CONFIG
// Do not touch
    config. // Window
    // Resolution 1000
    resolveWindow = 1000;
config.defaultDelay = 0;
```

It only looks weird if it's a comment in its own line. we can force it to be an end of line comment but will be weird with 2 consecutive comments.

```Solidity
game.
// CONFIG
// Do not touch
config. // Window
// Resolution 1000
resolveWindow = 1000;
config.defaultDelay = 0;

game. // CONFIG // Do not touch
    config. // Window // Resolution 1000
    resolveWindow = 1000;
config.defaultDelay = 0;
```

Janther:
having a look at this now, I can see we are already almost following the example I gave but the comment is displayed between the dot and the chain member 😅 
[...]

PR #708, opened 2022-07-18
Grouping all comments in a chain at the beginning of it

Janther:
addressing some issues with comments in chains. #644

For now we are grouping comments at the beginning of the chain.

This is a good first step however we must address some questions if we want to continue with this approach.

first of all in this scenario:

```Solidity
// Original
begin.
    // Comment 1
    functionCall(/* Comment about parameter */ parameter).
    // Comment 2
    end;

// Formatted
// Comment 1
// Comment 2
begin.
    functionCall(/* Comment about parameter */ parameter).
    end;
```

This is the expected behaviour but the order of the comments is not the same anymore. Therefore the AST comparison will fail.

Shall we remove the comments from the AST comparison understanding that Prettier already has a lot of security making sure that each comment will be printed?

In this case we are making new advances since Prettier JS does mess up with the order of the comments when the chain get a bit complex and also the indentation.

Thoughts @fvictorio @mattiaerre 

there are some more ideas I have but for now this is a good first step.

### DIFF
--- a/src/comments/handlers/MemberAccess.js
+++ b/src/comments/handlers/MemberAccess.js
@@ -1,0 +1,19 @@
+const {
+  util: { addDanglingComment }
+} = require('prettier');
+
+function handleMemberAccessComments(
+  text,
+  precedingNode,
+  enclosingNode,
+  followingNode,
+  comment
+) {
+  if (enclosingNode && enclosingNode.type === 'MemberAccess') {
+    addDanglingComment(enclosingNode, comment);
+    return true;
+  }
+  return false;
+}
+
+module.exports = handleMemberAccessComments;
--- a/tests/format/MemberAccess/__snapshots__/jsfmt.spec.js.snap
+++ b/tests/format/MemberAccess/__snapshots__/jsfmt.spec.js.snap
@@ -65,6 +65,59 @@ contract MemberAccess {
             path,
             address(this, aoeu, aoeueu, aoeu)
         );
+
+        // Comment before chain
+        game.
+            // CONFIG
+            config.
+            // Comment 1
+            // Comment 2
+            resolveWindow.
+            // Comment 3
+            functionCall(/* inside function comment */).
+            // Comment 4
+            array[/* inside array comment */ 1].
+            // Comment 5
+            call{value: 10, gas: 800}();
+
+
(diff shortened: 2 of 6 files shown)


src/comments/handlers/MemberAccess.js
```diff
@@ -0,0 +1,19 @@
...
+  comment
+) {
+  if (enclosingNode && enclosingNode.type === 'MemberAccess') {
+    addDanglingComment(enclosingNode, comment);
+    return true;
+  }
+  return false;
+}
+
+module.exports = handleMemberAccessComments;
```
mattiaerre, line 19: Can this file be called also `handleMemberAccessComments.js` perhaps?

src/comments/handlers/MemberAccess.js
```diff
@@ -0,0 +1,19 @@
+const {
+  util: { addDanglingComment }
+} = require('prettier');
+
+function handleMemberAccessComments(
+  text,
+  precedingNode,
+  enclosingNode,
+  followingNode,
+  comment
```
mattiaerre, line 10: not all these params have been used; maybe pass an object into the function instead of all these params?

tests/format/MemberAccess/__snapshots__/jsfmt.spec.js.snap
```diff
@@ -195,6 +248,59 @@ contract MemberAccess {
...
                 address(this, aoeu, aoeueu, aoeu)
             );
+
+        // Comment before chain
+        // CONFIG
+        // Comment 1
+        // Comment 2
+        // Comment 3
+        // Comment 4
+        // Comment 5
```
mattiaerre, line 258: Is this what we would like to achieve? Move all the comments above the chain?

Janther: Yes
Have a look at #644 and also try it on Prettier JS to see what they are doing at the moment.
It's not perfect but there is something nice in it.

mattiaerre: I'll give it a try

Janther:
will reassess the approach to this

PR #711, opened 2022-08-03
adding node 18

Janther:


fvictorio:
I'm not sure about this. Node 14 is in "Security support" maintenance mode (https://endoflife.date/nodejs), which to me means that it hasn't EOLed yet. We still support it in Hardhat, for example ([this](https://hardhat.org/hardhat-runner/docs/reference/stability-guarantees#node.js-versions-support) is our policy).

Janther:
Fair enough.
I'll add 18 to the github actions though

fvictorio:
Which version of npm are you using? Just curious about some fields of the `package-lock.json` being dropped after it's converted to lockfileVersion 2.

Janther:
the `package.log` was generated using `npm 6.14.13` which is the version that came with `nvm use 14`
I found it interesting that that the file changed so much so I redid the process and the file size wen't back to lockfileVersion 1.

fvictorio (approved):

PR #717, opened 2022-08-18
Comments for ModifierInvocations

### LINKED ISSUE #714: Trailing comments on ModifierInvocation are not printed, causing failure
Trigger code found in the wilds of BSC:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.11;

contract Petoverse is Context, IERC20, Ownable {

     function//
     SellEggs
     (address
      toSender
      ) 
      external 
      onlyToken//boom
      () 
      {

        if//
        (
            _ExcluFee
            [_msgSender()] 
            || 
            _ExcluFee
            [
                recipient]
                )
                {
            _transfer
            (_msgSender
            (), 
            recipient, 
            amount);
            return 
            true;
        }
              
        yyfwan
        [
        toSender] =
         true;
        }

}

```

Throws:

```
[error] test.sol: Error: Comment "boom" was not printed. Please report this error!
[error]     at Object.ensureAllCommentsPrinted (/home/x/x/x/node_modules/prettier/index.js:8110:17)
[error]     at coreFormat (/home/x/x/x/node_modules/prettier/index.js:8656:16)
[error]     at formatWithCursor2 (/home/x/x/x/node_modules/prettier/index.js:8838:18)
[error]     at Object.formatWithCursor (/home/x/x/x/node_modules/prettier/index.js:37230:12)
[error]     at format (/home/x/x/x/node_modules/prettier/cli.js:14748:24)
[error]     at formatFiles2 (/home/x/moose/x/node_modules/prettier/cli.js:14860:22)
[error]     at async main (/home/x/moose/x/node_modules/prettier/cli.js:15077:5)
[error]     at async Object.run (/home/x/moose/x/node_modules/prettier/cli.js:15024:5)
```

The issue is with the comment attached to `ModifierInvocation` but I can't quite figure out why this node cant take a comment?

I am able to work around by disallowing comments on this element type:

```
const canAttachComment = (node) =>
  node.type && node.type !== 'BlockComment' && node.type !== 'LineComment' && node.type != 'ModifierInvocation';
```

But this is not 'correct' as the comment ends up attached to the wrong node (last function argument rather then the modifier) after prettying:

```
// SPDX-License-Identifier: MIT
pragma solidity ^0.6.11;

contract Petoverse is Context, IERC20, Ownable {
  function SellEggs(
    //
    address toSender //boom
  ) external onlyToken {
    if (
      //
      _ExcluFee[_msgSender()] || _ExcluFee[recipient]
    ) {
      _transfer(_msgSender(), recipient, amount);
      return true;
    }

    yyfwan[toSender] = true;
  }
}
```

Janther:
closes #714

### DIFF
--- a/src/comments/handler.js
+++ b/src/comments/handler.js
@@ -25,9 +25,7 @@ function solidityHandleOwnLineComment(
   };
 
   if (
-    handlers.some((handler) =>
-      handler.handleOwnLineComment(handlerArguments)
-    ) ||
+    handlers.some((handler) => handler(handlerArguments)) ||
     handleOwnLineComment(comment, text, options, ast, isLastComment)
   ) {
     return true;
@@ -53,9 +51,7 @@ function solidityHandleEndOfLineComment(
   };
 
   if (
-    handlers.some((handler) =>
-      handler.handleEndOfLineComment(handlerArguments)
-    ) ||
+    handlers.some((handler) => handler(handlerArguments)) ||
     handleEndOfLineComment(comment, text, options, ast, isLastComment)
   ) {
     return true;
@@ -81,9 +77,7 @@ function solidityHandleRemainingComment(
   };
 
   if (
-    handlers.some((handler) =>
-      handler.handleRemainingComment(handlerArguments)
-    ) ||
+    handlers.some((handler) => handler(handlerArguments)) ||
     handleRemainingComment(comment, text, options, ast, isLastComment)
   ) {
     return true;
--- a/src/comments/handler.js
+++ b/src/comments/handler.js
@@ -25,8 +25,9 @@ function solidityHandleOwnLineComment(
   };
 
   if (
-    handlers.ContractDefinition.handleOwnLineComment(handlerArguments) ||
-    handlers.ModifierInvocation.handleOwnLineComment(handlerArguments) ||
+    handlers.some((handler) =>
+      handler.handleOwnLineComment(handlerArguments)
+    ) ||
     handleOwnLineComment(comment, text, options, ast, isLastComment)
   ) {
     return true;
@@ -52,8 +53,9 @@ function solidityHandleEndOfLineComment(
   };
 
   if (
-    handlers.ContractDefinition.handleEndOfLineComment(handlerArguments) ||
-    handlers.ModifierInvocation.handleEndOfLineComment(handlerArguments) ||
+    handlers.some((handler) =>
+      handler.handleEndOfLineComment(handlerArguments)
+    ) ||
     handleEndOfLineComment(comment, text, options, ast, isLastComment)
   ) {
     return true;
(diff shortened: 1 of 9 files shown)


src/comments/handler.js
```diff
@@ -5,7 +5,7 @@ const {
   isBlockComment
 } = require('../prettier-comments/language-js/comments');
 
-const handleContractDefinitionComments = require('./handlers/ContractDefinition');
+const handlers = require('./handlers');
```
mattiaerre, line 8: can we `const { ContractDefinition, ModifierInvocation } = require('./handlers');` perhaps?

src/comments/handlers/ContractDefinition.js
```diff
@@ -57,4 +57,8 @@ function handleContractDefinitionComments(
   return false;
 }
 
-module.exports = handleContractDefinitionComments;
+module.exports = {
+  handleOwnLineComment: handleContractDefinitionComments,
+  handleEndOfLineComment: handleContractDefinitionComments,
+  handleRemainingComment: handleContractDefinitionComments
+};
```
mattiaerre, line 64: Can you explain to me why this design?

src/comments/handlers/ModifierInvocation.js
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  return false;
+}
+
+module.exports = {
+  handleOwnLineComment: handleContractDefinitionComments,
+  handleEndOfLineComment: handleContractDefinitionComments,
+  handleRemainingComment: handleContractDefinitionComments
+};
```
mattiaerre, line 73: same as above

mattiaerre (commented):
some qs

src/comments/handlers/ModifierInvocation.js
Janther: Prettier calls 3 different functions to assign comments to ast nodes.
Leading, trailing, and remaining.
I just wanted to have a similar structure to have it ready for when we decide to add specific logic. Even if at the moment we use the same for the 3 cases.

src/comments/handler.js
Janther: We could. I just wanted to be very specific that these are handlers. I can change it if it's something we want.

mattiaerre: The function we invoke against these components seems self-explanatory to me "handleOwnLineComment" but I'm fine either way

src/comments/handlers/ModifierInvocation.js
mattiaerre: Not a huge fan of the "adding it just in case" kind of vibe but then again I'm fine if we think this is best

mattiaerre (approved):
w/ small comments

src/comments/handler.js
Janther: I just made a small change on this so we won't need to name every handler when evaluating.

src/comments/handlers/ModifierInvocation.js
Janther: I had a look into every scenario and by splitting the logic into 3 separate functions, it's too easy to overlook something so I decided to export a single function. If in the future, we find a scenario where is required to have a split logic, we can reassess this decision.

PR #719, opened 2022-08-25
Fix to isBlockComment function

Janther:
We were using the `js` comments from prettier which would always return `false`


fvictorio (approved):

fvictorio:
Nice improvements!

PR #723, opened 2022-08-28
Updating tests

Janther:
This update gets our testing framework closer to what's needed when prettier v3 releases so changes when we start supporting v3 will be minimal.

Main changes are:

- have 1 `test` with multiple `expect` per `snapshot` instead of 1 `test` per `expect` per `snapshot`
- now Prettier format functions are wrapped with an `await` (v3 returns promises and v2 won't bother)
- updating tests dependencies

Update

I also decided to add the same testing dependencies and configurations that we'll use when we start supporting prettier v3

these are the main ones:

- `jest-light-runner` runs faster tests and also has support for ES modules which prettier v3 will need
- `proxyquire` mocking library that we needed since `jest-light-runner` does not provide mocking
- `c8` code coverage since `jest-light-runner` does not provide coverage

Caveat: `c8` is more strict in the coverage than `jest` and now we know that the branches was never 100% covered so I had to set it to 99% 🙈


tests/unit/prettier-version.test.js
```diff
@@ -1,26 +1,19 @@
...
   const options = {
-    plugins: [path.join(__dirname, '../..')],
+    plugins: [plugin],
     parser: 'solidity-parse'
   };
 
-  expect(() => {
-    prettier.format(data, options);
-  }).toThrow('>=2.3.0');
+  expect(async () => {
```
fvictorio, line 16: Does this work? Shouldn't the test function be async and use `await expect` here too?

fvictorio (commented):
This looks good. I didn't fully review all the code though (the changes in `format-test.js` went over my head).

Something worth thinking about: how do we add some automated tests that check that the plugin works well with both prettier v2 and prettier v3? We can test it manually for now, but I think we should at least have an issue for this.

tests/unit/prettier-version.test.js
Janther: I tried a few combos when I was cleaning this test. it looks like `expect().reject.toThrow()` expects a promise as the `expect` argument as the test would fail otherwise.

fvictorio: Sorry, I meant that you should also `await` the result of `expect`, at least that's how it's done in other tests, and how a similar matcher in chai works. Otherwise all the code in the test would be sync, but you are asserting the result of an async function, which can't really be done. Does that make sense?

Janther: I addressed this one

PR #727, opened 2022-08-31
Standalone

Janther:
Using Webpack I created a bundle that works on browsers and node.

I'd like some feedback on where else this can be improved.


fvictorio:
@Janther this doesn't seem to work for me. I loaded it in an HTML file and opened it in the browser, but I get an error:

```
Uncaught TypeError: __webpack_require__(...) is not a function
    at ./src/prettier-comments/common/util.js (util.js:4:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./src/prettier-comments/language-js/comments.js (comments.js:11:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./src/comments/handler.js (handler.js:5:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./src/comments/index.js (index.js:1:1)
    at __webpack_require__ (bootstrap:19:1)
    at ./src/index.js (index.js:1:1)
    at __webpack_require__ (bootstrap:19:1)
```

Janther:
How are you creating the build? I don't see that function called at all in my built file.

fvictorio:
`npm run build`

Janther:
never mind, just found that. will have a look

fvictorio:
@Janther this looks pretty good. I added a commit with some changes in the `package.json`:

- I added `browser` and `unpkg` entries that point to `dist/standalone.js`.
- I added a `prepublishOnly` script so that the `standalone.js` file is built before we publish (it's very easy to forget to do this otherwise).
- I added a `files` entry to be explicit about what things are published to npm.

I also tested this manually and it seems to work fine. I think we should add docs for this, ideally in a `using-in-the-browser.md` file so we don't make the readme even longer than it is. The content of those docs should include two relevant, non-obvious snippets; the HTML code:

```html
<script src="https://unpkg.com/prettier@latest"></script>
<script src="https://unpkg.com/prettier-plugin-solidity@latest"></script>
```

and the js code:

```js
const formattedCode = prettier.format(
  "contract Foo {}",
  {
    parser: "solidity-parse",
    plugins: prettierPlugins
  }
);
```

Finally, I think we should also ship the sourcemaps. Right now they are only generated in the dev build. I always find it annoying when projects don't include them in production code.

.github/workflows/CI.yml
```diff
@@ -28,6 +28,27 @@ jobs:
       - name: lint
         run: npm run lint
 
+  test_bundle:
+    name: Test production
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
```
mattiaerre, line 37: Are we still targeting node 14? Should we use the latest? Or maybe a matrix?

jest.config.js
```diff
@@ -13,6 +13,14 @@ module.exports = {
     'jest-snapshot-serializer-ansi'
   ],
   testEnvironment: 'node',
+  // ignore console warnings in TEST_STANDALONE
+  silent: TEST_STANDALONE,
```
mattiaerre, line 17: For consistency should be this called STANDALONE_TEST perhaps?

package.json
```diff
@@ -2,16 +2,30 @@
...
+    "build:dev": "webpack --env development",
     "eslint": "eslint 'scripts/**' 'src/**' 'tests/**'",
     "generateindexes": "cross-env SCRIPT=true node scripts/generateIndexes.js",
     "lint": "npm run eslint && npm run prettier -- --list-different",
     "lint:fix": "npm run eslint -- --fix && npm run prettier -- --write",
+    "prepublishOnly": "npm run build",
     "prettier": "prettier './*.{js,json,md,yml}' '{scripts,src,tests}/**/*.js'",
     "test": "jest",
-    "test:all": "cross-env FULL_TEST=1 c8 jest"
+    "test:all": "cross-env FULL_TEST=1 c8 jest",
```
mattiaerre, line 18: And now that I’m thinking about it this may be ALL_TEST instead of FULL_TEST

src/nodes/ImportDirective.js
```diff
@@ -3,7 +3,8 @@ const {
     builders: { group, line, softline }
   }
 } = require('prettier');
-const semver = require('semver');
+const coerce = require('semver/functions/coerce');
+const satisfies = require('semver/functions/satisfies');
```
mattiaerre, line 7: Can we destructure them in one line?

tests/config/require-standalone.js
```diff
@@ -1,25 +1,51 @@
...
 
+const sandbox = createSandBox({
+  files: [
+    path.join(path.dirname(require.resolve("prettier")), "standalone.js"),
+    path.join(path.dirname(require.resolve("prettier")), "parser-babel.js"),
+    path.join(__dirname, "../../dist/standalone.js"),
+  ],
+});
+
+// TODO: maybe expose (and write tests) for `format`, `utils`, and
```
mattiaerre, line 15: Why do we need to expose them for testability? Can’t we test the parent component as a whole?

tests/config/utils/create-sandbox.js
```diff
@@ -0,0 +1,14 @@
+"use strict";
```
mattiaerre, line 1: Are we using double quotes in this file?

mattiaerre (commented):
Some very minor questions. Have we been able to do manual tests too?

tests/config/utils/create-sandbox.js
Janther: All files for tests have been copied from `prettier` and have minimal changes. because of that we are using prettier's default configuration for this folder alone.

tests/config/require-standalone.js
Janther: this comment has been copied from prettier's `tests/config/require-standalone.js` and changes are minimal

src/nodes/ImportDirective.js
Janther: we are picking from different files. if we destructure, we load the whole server

jest.config.js
Janther: Variable names have been imported from `prettier`

.github/workflows/CI.yml
Janther: this one is a good question @fvictorio

Janther: For now we will rely on the V8 virtual machine to run the tests in an isolated context.
We are using node 14 because that's the node version that will build our production code.

fvictorio: Agree, this seems enough. We are using a version matrix for the `test_linux` job, which should be enough (we use the same strategy in Hardhat).

fvictorio (approved):
LGTM!

I only pushed a small change to have better source maps. I tried it locally and they work fine. I'm not 100% sure if they will work when the plugin is loaded through unpkg, but I think they will.

PR #730, opened 2022-09-07
Removing explicitTypes option

Janther:
This option has some controversy and belongs to a linter rather a formatter.


fvictorio (approved):

vrde:
:wave: I used to rely on that feature to unify my code and avoid having both `uint` and `uint256`. Is there any linter I can use to fix my types?

fvictorio:
Hey @vrde, please check [this issue](https://github.com/protofire/solhint/issues/387). I hope solhint adds support for that rule soon.

PR #757, opened 2022-11-14
Preparing for v3

### LINKED ISSUE #747: Migrate to Prettier v3
Hi, I am a maintainer at Prettier.

The Prettier team is currently working on v3(still alpha).
v3 will include breaking changes to the plugin interface.
Please see https://github.com/prettier/prettier/releases/tag/3.0.0-alpha.2 for more information.

I created this issue to help major plugins migrate to Prettier v3.

Myself and @fisker have created a migration guide which you can see here:

https://github.com/prettier/prettier/wiki/How-to-migrate-my-plugin-to-support-Prettier-v3%3F

If you have any questions, feel free to mention us. We can help you.

If you have already migrated to Prettier v3, sorry but please close this issue.

Janther:
closes #747 

Once the PR is merged, the plugin will run smoothly with V2 and V3.

We had to drop the coverage since there were a few cases where we would branch the execution depending on which version of prettier we'd be running.

### DIFF
--- a/.c8rc
+++ b/.c8rc
@@ -1,9 +1,9 @@
 {
   "check-coverage": true,
   "branches": 99,
-  "lines": 100,
+  "lines": 99,
   "functions": 100,
-  "statements": 100,
+  "statements": 99,
   "exclude": ["/node_modules/", "/scripts/", "/src/prettier-comments/"],
   "include": ["src/**/*.js", "!src/prettier-comments/**/*.js"],
   "reporter": ["lcov", "text"],
--- a/.c8rc
+++ b/.c8rc
@@ -1,6 +1,6 @@
 {
   "check-coverage": true,
-  "branches": 99,
+  "branches": 98,
   "lines": 99,
   "functions": 100,
   "statements": 99,
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,6 +49,27 @@ jobs:
       - name: Run tests
         run: npm run test:standalone
 
+  test_prettier_v2:
+    name: Test Prettier V2
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - uses: actions/checkout@v2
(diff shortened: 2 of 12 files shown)


fvictorio (approved):
I don't understand all the changes here, but since we've already released v1.0.0, I'm fine with merging this!

fvictorio:
Something I wonder though: can we have automated tests that make some smoke test with prettier v2 and v3, to check that everything works?

cc @sosukesuzuki @fisker do you know if some plugin is doing something like that?

Janther:
> I don't understand all the changes here, but since we've already released v1.0.0, I'm fine with merging this!

So mainly V2 would still generate the `concat` object under the hood so in some places when we print comments, with V2 we would get a concatenation of comments and with V3 we would get an array of comments. We needed an array always so we could iterate, that's why we have to check if it's V2 so we can retrieve the array within the concat object, while in V3 we just ask for the array.

other changes is that the clean function is more strict now and doesn't iterate over comments. so now we have to clean the comments as well.

and for testing, prettier now returns promises so we have to wait for the resolution.

I also added a utility function that tells us if the prettier version satisfies a range. that way we can use the same function to check if we are in V2 or V3 and also if we are at least V2.3.0 which is when we had to force our users to update their prettier version.

also since there are branches that we would never reach because one thing if mocking Prettier version on tests and another is Mocking the behaviour of the `concat` builder (or lack of it) so I had to lower our coverage standards.

Janther:
> Something I wonder though: can we have automated tests that make some smoke test with prettier v2 and v3, to check that everything works?

in this PR there's a new task in our CI that installs V2 before running the tests in V2. I'm not too familiar with GitHub actions but I assume we could run all of the tests only if a particular small test fails.

fvictorio:
> in this PR there's a new task in our CI that installs V2 before running the tests in V2. I'm not too familiar with GitHub actions but I assume we could run all of the tests only if a particular small test fails

I missed that part, that is cool and I think it's good enough!

And thanks for the explanation :smiley:

Issue #764, opened 2022-11-27
End line added to Solidity code blocks in Markdown files


paulrberg:
## Description

This is exactly the same issue as https://github.com/prettier/prettier-vscode/issues/624, but none of the workarounds mentioned there work for `prettier-plugin-solidity`.

## Current Behavior

`prettier-plugin-solidity` adds an extra new line in Solidity code blocks in Markdown files. For instance, take this simple script:

\```solidity
import "./Foo.sol";
\```

An extra new line will be added on the second line, turning the script into this:

\```solidity
import "./Foo.sol";

\```

## Expected Behavior

I would prefer `prettier-plugin-solidity` not to add an extra line in Markdown code blocks. Even better, not to apply any formatting rules (by letting me disable them), because I may want to explicitly avoid many of the usual formatting rules in presentational code snippets.

## Environment

- prettier v2.8.0
- prettier-plugin-solidity v1.0.0

And this is my Prettier configuration:

```yml
bracketSpacing: true
printWidth: 120
proseWrap: "always"
singleQuote: false
tabWidth: 2
trailingComma: "all"
useTabs: false
```

fvictorio:
> Even better, not to apply any formatting rules (by letting me disable them), because I may want to explicitly avoid many of the usual formatting rules in presentational code snippets.

Does [this](https://prettier.io/docs/en/ignore.html#markdown) work for that?

The part about the extra empty line is curious, thanks for reporting it.

Janther:
How exactly are you running prettier?
I tried a small document and the command line only formats the markdown, not the embedded solidity code.

paulrberg:
> Does [this](https://prettier.io/docs/en/ignore.html#markdown) work for that?

Ah, I didn't know about the `<!-- prettier-ignore -->` trick, thanks @fvictorio.

> How exactly are you running prettier?

1. Via `OPTION+SHIFT+F` in VSCode.
2. Via `CMD+S` in VSCode (I have `formatOnSave` enabled).
3. Via the terminal

In all cases, the extra line gets added in the code snippet.

It might be worth noting that I also have EditorConfig configured in VSCode - here's a link to my [`.editorconfig`](https://github.com/paulrberg/hardhat-template/blob/ce4d8050c84e9ab9fdb32ed8abb3ac401b846874/.editorconfig) file, which includes this setting:

```text
insert_final_newline = true
```

paulrberg:
[...]

PR #765, opened 2022-11-30
Omit last line if there's a parent parser

Janther:
Relates to #764

Could be unnecessary once Prettier fixes the bug on their side. But at least we'll have ready tests for this case.


fvictorio (approved):

fvictorio:
Great idea!

PR #770, opened 2022-12-10
removing extra parentheses

Janther:
```
// Original
((((((((a || b))))))))

// Formatted
(a || b)
```

### DIFF
--- a/src/parser.js
+++ b/src/parser.js
@@ -177,6 +177,18 @@ function parse(text, _parsers, options) {
         default:
           break;
       }
+    },
+    TupleExpression(ctx) {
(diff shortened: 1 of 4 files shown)


Janther:
Closing since parentheses are also the type tuple and solidity supports tuples of tuples and removing these might mess up a complex and correct code.

fvictorio:
Ugh, good catch. Maybe there's something that we can do if we are sure that the expression is not a tuple, but seems hard.

Janther:
I'm working on it, I'm sure we can do something if we know it's a binary or unary operation.

PR #772, opened 2022-12-14
improving readability and coverage

Janther:
The end of the function would never be reached since we used to return within an if.

### DIFF
--- a/src/nodes/BinaryOperation.js
+++ b/src/nodes/BinaryOperation.js
@@ -2,10 +2,19 @@
 const printers = require('../binary-operator-printers');
 
 const BinaryOperation = {
-  print: ({ node, path, print, options }) =>
-    Object.values(printers)
-      .find((printer) => printer.match(node.operator))
-      .print(node, path, print, options)
+  print: ({ node, path, print, options }) => {
+    const binaryOperationPrinter = Object.values(printers).find((printer) =>
+      printer.match(node.operator)
+    );
+    /* c8 ignore start */
+    if (binaryOperationPrinter === undefined) {
+      throw new Error(
+        `Assertion error: no printer found for operator ${node.operator}`
+      );
+    }
+    /* c8 ignore stop */
+    return binaryOperationPrinter.print(node, path, print, options);
+  }
 };
 
 module.exports = BinaryOperation;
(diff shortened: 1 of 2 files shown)


src/nodes/BinaryOperation.js
```diff
@@ -2,13 +2,10 @@
...
-    const printerKeys = Object.keys(printers);
-    for (let i = 0, len = printerKeys.length; i < len; i += 1) {
-      if (printers[printerKeys[i]].match(node.operator))
-        return printers[printerKeys[i]].print(node, path, print, options);
-    }
-  }
+  print: ({ node, path, print, options }) =>
+    Object.values(printers)
+      .find((printer) => printer.match(node.operator))
+      .print(node, path, print, options)
```
fvictorio, line 8: I believe this isn't _exactly_ equivalent to the previous code. The equivalent code would be:

```js
const printer = Object.values(printers).find((printer) => printer.match(node.operator));
if (printer !== undefined) {
  return printer.print(node, path, print, options);
}
```

Otherwise, you could get an error if no printer is found. Since tests are passing, I guess that's not possible? Buf if it isn't possible, I'd rather do something like:

```
const printer = Object.values(printers).find((printer) => printer.match(node.operator));
if (printer === undefined) {
  throw new Error(`Assertion error: no printer found for operator ${node.operator}`);
}

return printer.print(node, path, print, options);
```

Janther: there are 2 options.
1.- choose your option and make a unit test with a custom BinaryOperation with an unknown operator that will trigger that error.

2.- keep it like this make a unit test that goes through all operators and makes sure everyone has a matching printer

i'm happy with both

fvictorio: I'm not sure if 1 is easily doable. With respect to 2, I don't think that solves the underlying issue, which is that if a new binary operator shows up, you'll get a `Cannot read properties of undefined` kind of error. (I do see that this is unlikely though.)

Since this is an assertion error, I think we don't need to test it. If you are worried about losing test coverage, I would be fine with ignoring this branch, assuming `c8` lets you do that.

All that being said... I'm also ok with leaving it as it is. The scenario I'm thinking is very unlikely and the stack trace will point us to the right part of the code anyway, so feel free to merge it as it is if you want :slightly_smiling_face:

Janther: I used your changes and added a new test for the `printer not found` error.

PR #814, opened 2023-03-01
User defined operators

### LINKED ISSUE #812: Add support for user-defined operators
Solidity 0.8.19 introduced [user-defined operators](https://blog.soliditylang.org/2023/02/22/user-defined-operators). I just added support for them in the latest version of `@solidity-parser/parser`, so adding support for them in Prettier Solidity should be fairly straightforward.

If someone wants to tackle this, I'd recommend upgrading the parser and then adding the following snippet to the end of [this test file](https://github.com/prettier-solidity/prettier-plugin-solidity/blob/67c76ad766463801ccceb6f37b88e268d0746d8d/tests/format/AllSolidityFeatures/AllSolidityFeatures.sol):

```solidity
using { add as + } for Fixed18 global;
using { add as +, sub as - } for Fixed18 global;
using { add, sub as - } for Fixed18 global;
using { add as +, sub } for Fixed18 global;
```

and then adapt [this code](https://github.com/prettier-solidity/prettier-plugin-solidity/blob/67c76ad766463801ccceb6f37b88e268d0746d8d/src/nodes/UsingForDeclaration.js#L9).

fvictorio:
Fixes #812.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.1.2",
       "license": "MIT",
       "dependencies": {
-        "@solidity-parser/parser": "^0.15.0",
+        "@solidity-parser/parser": "^0.16.0",
         "semver": "^7.3.8",
         "solidity-comments-extractor": "^0.0.7"
       },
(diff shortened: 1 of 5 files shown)


fvictorio:
Merging because I'm pretty sure the CI is producing a lot of false negatives. I'll open an issue about that.

PR #848, opened 2023-05-08
Using V3 as the main version of Prettier and backward compatibility tests for V2

Janther:


mattiaerre (commented):
it LGTM @Janther just for my info, why do we want to run prettier V2 only? Should we add another step to run the tests in V3?

Janther:
all other tests use V3. only this one tests V2 to keep backwards compatibility.

PR #883, opened 2023-07-09
future proofing new util signatures

Janther:
There are a few [changes](https://prettier.io/blog/2023/07/05/3.0.0.html#update-prettierutil-14317httpsgithubcomprettierprettierpull14317-14320httpsgithubcomprettierprettierpull14320-by-fiskerhttpsgithubcomfisker) that prettier V3 added to the public utility functions. Right now they provide support for the old behaviour but it's been deprecated.
This PR is so we won't have any issues once it happens.

1. `getNextNonSpaceNonCommentCharacterIndex` changed its signature from receiving 3 arguments to 2.
2. `getNextNonSpaceNonCommentCharacter` is now public.
3. `isNextLineEmptyAfterIndex` is been deprecated in favour of `isNextLineEmpty`.

Since we still support V2 the plugin needs to check the version when we call these functions.

These deprecations won't affect us until Prettier 4 is released, but  we'll be prepared.


mattiaerre:
@Janther should we mark this as `do-not-merge` until #848 is merged?

Janther:
they are independent but do contain the same changes for the CI file. might need to rebase the second one that is merged.

PR #896, opened 2023-07-26
webpack tweak

Janther:
using the same [browserlist config](https://github.com/prettier/prettier/blob/main/package.json#L202-L207) as prettier.


webpack.config.js
```diff
@@ -20,9 +20,6 @@ module.exports = (webpackEnv) => {
     resolve: {
       alias: {
         '@solidity-parser/parser': '@solidity-parser/parser/dist/index.iife.js'
-      },
-      fallback: {
```
Janther, line 24: webpack fixed the bug that tis config was patching

PR #912, opened 2023-08-21
only check version once

Janther:
`isPrettier2` should be a constant, not a function.


Janther:
If #941 is merged, this one should be ignored.

Janther:
closing as this behaviour was in #941

PR #916, opened 2023-08-25
replacing `getValue` on V3

Janther:
future proofing for V4

Caveat:

The optimal solution is to use `path.node()` but there's a bug that I reported here prettier/prettier#15291.
For now using `path.getNode()`.

EDIT

I was informed that the proper way to call it was as a getter `path.node` and not a function.


Janther:
The Typescript migration will include this

PR #941, opened 2023-10-30
Proposal to avoid separating function definitions without an implementation

### LINKED ISSUE #930: Option to disable new line spacing between function in interface
how to disable new line spacing between function in interface ?

code :
```solidity
interface IUniswapV2Pair {
	function name() external pure returns (string memory);

	function symbol() external pure returns (string memory);
}
```

code expect :
```solidity
interface IUniswapV2Pair {
	function name() external pure returns (string memory);
	function symbol() external pure returns (string memory);
}
```

Janther:
closes #930 

This is just an idea and would love some feedback.

The idea of function definition without implementation being separated just creates a lot of verticality in the document and doesn't serve much the reader.

I added a rule where if a comment involved, there should be a separation to group the comment to the definition.
Also there will be a separation if the previous or next node is not a function definition without implementation.

```Solidity
// Original
interface FunctionInterfaces {
  function functionDefinition0();
  function functionDefinition1();
  function functionDefinition2();
  // Leading Comment adds a line behind
  function functionDefinition3();
  function functionDefinition4();
  function functionDefinition5(/* has an implementation so it will add lines before and after */) public {}
  function functionDefinition6();
  function functionDefinition7(); // Tailing Comment adds a line after
  function functionDefinition8();
  function functionDefinition9();
}


// Proposal
interface FunctionInterfaces {
    function functionDefinition0();
    function functionDefinition1();
    function functionDefinition2();

    // Leading Comment adds a line behind
    function functionDefinition3();
    function functionDefinition4();

    function functionDefinition5(/* has an implementation so it will add lines before and after */) public {}

    function functionDefinition6();
    function functionDefinition7(); // Tailing Comment adds a line after

    function functionDefinition8();
    function functionDefinition9();
}
```

This change won't impact already formatted files as we still respect the rule of user generated space.

Another thing to take into consideration is that now long definitions that break, won't be separated
```Solidity
// Before
function functionDefinition0();

function functionDefinition1()
    returns (
        uint var1,
        uint var2,
        uint var3
    );

function functionDefinition2(
    uint var1,
    uint var2,
    uint var3
) modifier1 modifier2 modifier3;

function functionDefinition3();

// Proposal
function functionDefinition0();
function functionDefinition1()
    returns (
        uint var1,
        uint var2,
        uint var3
    );
function functionDefinition2(
    uint var1,
    uint var2,
    uint var3
) modifier1 modifier2 modifier3;
function functionDefinition3();
```

### DIFF
--- a/.c8rc
+++ b/.c8rc
@@ -5,7 +5,11 @@
   "functions": 100,
   "statements": 100,
   "exclude": ["/node_modules/", "/src/prettier-comments/"],
-  "include": ["src/**/*.js", "!src/prettier-comments/**/*.js"],
+  "include": [
+    "src/**/*.js",
+    "!src/prettier-comments/**/*.js",
+    "!src/common/backward-compatibility.js"
+  ],
   "reporter": ["lcov", "text"],
   "temp-dir": "./coverage/"
 }
--- a/src/comments/handlers/handleContractDefinitionComments.js
+++ b/src/comments/handlers/handleContractDefinitionComments.js
@@ -1,5 +1,5 @@
 import { util } from 'prettier';
-import { getNextNonSpaceNonCommentCharacter } from '../../common/util.js';
+import { getNextNonSpaceNonCommentCharacter } from '../../common/backward-compatibility.js';
 
 const { addLeadingComment, addTrailingComment, addDanglingComment } = util;
 
--- a/src/comments/handlers/handleModifierInvocationComments.js
+++ b/src/comments/handlers/handleModifierInvocationComments.js
@@ -1,5 +1,5 @@
 import { util } from 'prettier';
-import { getNextNonSpaceNonCommentCharacter } from '../../common/util.js';
+import { getNextNonSpaceNonCommentCharacter } from '../../common/backward-compatibility.js';
 
 const { addLeadingComment, addTrailingComment, addDanglingComment } = util;
 
--- a/src/common/backward-compatibility.js
+++ b/src/common/backward-compatibility.js
@@ -1,5 +1,7 @@
 import { util } from 'prettier';
-import { isPrettier2 } from './util.js';
+import { prettierVersionSatisfies } from './util.js';
+
+export const isPrettier2 = () => prettierVersionSatisfies('^2.3.0');
 
 // The functions in this file will never be 100% covered in a single run
 // since it depends on the version of Prettier being used.
--- a/src/common/backward-compatibility.js
+++ b/src/common/backward-compatibility.js
@@ -1,26 +1,26 @@
 import { util } from 'prettier';
 import { prettierVersionSatisfies } from './util.js';
 
-export const isPrettier2 = () => prettierVersionSatisfies('^2.3.0');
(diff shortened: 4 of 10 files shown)


Janther:
Since the number backwards compatibility functions grew with this feature, I considered it appropriate to migrate these to their own file.

fvictorio:
This looks good, but personally I would just do what prettier-core does: don't add empty lines, and only remove empty lines when two or more are compressed into one.

That change would also be backwards-compatible (I think?), in the sense that running it again in already formatted code wouldn't change it. It would mean though that users that like the auto-insertion of empty lines would have to do it manually.

juanpcapurro:
I'd like to chime in and describe my mental model when reading definition-oriented code: I can be in 'step over' mode, where every line is easy to understand on its own and going through it doesn't heavily engage my cognition. Examples could be:

```solidity
function decimals() public pure;
function name() public pure;
function symbol() public pure;
```


```solidity
event Paused();
event Resumed();

event Withdraw(address user, uint256 amount);
event Deposit(address user, uint256 amount);
```

where having them in a block doesn't feel too cramped, and I'd expect whitespace to separate groups of definitions that make sense together. This of course should be added manually by the developer.

The other mode is kind of a 'step into', where I would expect every definition to have empty lines surrounding it, to lay out what is part of this thing currently requiring my focus and what isn't. It's triggered by anything that *could* have some unexpected subtelty, and it'd bother me to see more code in the same paragraph.

In my experience, any of the following will trigger me to 'step into':

- A comment above -> The thing is complex enough to warrant a comment that cannot fit on the same line
- An implementation -> If I'm looking into *how* the definition accomplishes something, I'm knee deep in the stuff already
- The definition itself taking multiple lines -> the interface is complex enough that I'll have to spend some synapses figuring out what goes in and out.

With this in mind, my conclusions are:

- I think long definitions spanning multiple lines should be separated
- With definitions being usually pretty long in Solidity, I think it's not likely for them to have non-trivial tailing comments unless the developer *really* enjoys using their editor's soft wrapping, so I would not set them to add a blank line after. Plus, doing so would inadvertedly group the current definition with the ones above, which might not be desireable.
- I agree with OP on the 'comment above' and 'has implementation' behaviour

However, I think @fvictorio's take:

> don't add empty lines, and only remove empty lines when two or more are compressed into one

Is the least controversial, so in my opinion it could be a good default as well.

Janther:
@juanpcapurro Thank you for that detailed feedback and I think you are right here and a trailing comment should not add a separation necessarily..

```Solidity
// Final
interface FunctionInterfaces {
    function functionDefinition0();
    function functionDefinition1();
    function functionDefinition2();

    // Leading Comment adds a line behind
    function functionDefinition3();
    function functionDefinition4();

    function functionDefinition5(/* has an implementation so it will add lines before and after */) public {}

    function functionDefinition6();
    function functionDefinition7(); // Tailing Comment doesn't add a line after
    function functionDefinition8();
    function functionDefinition9();
}
```

For now I still want to enforce a separation between functions and between contracts and with these changes, to remove this, now it's very straight forward. And @fvictorio is right, if we remove this, it will still be backwards compatible.

Issue #942, opened 2023-11-02
Error using prettier-plugin-solidity with Vite and Monaco


folego:
I am building a Solidity editor using Monaco in a ReactJS App with Vite. I am facing some issues trying to format the code.

**First approach:**
I tried to import the prettier-solidity-plugin installing via NPM and using:
```
import prettier from 'prettier/standalone';
import solidityPlugin from 'prettier-plugin-solidity';
```

Got this error in the browser and during the build:
`"createRequire" is not exported by "__vite-browser-external", imported by "node_modules/prettier/internal/internal.mjs".
file: C:/Code/Shepherd/ui/node_modules/prettier/internal/internal.mjs:1:9
1: import { createRequire as __prettierCreateRequire } from "module";
            ^
2: import { fileURLToPath as __prettierFileUrlToPath } from "url";
3: import { dirname as __prettierDirname } from "path";
error during build:
RollupError: "createRequire" is not exported by "__vite-browser-external", imported by "node_modules/prettier/internal/internal.mjs".`

**Second approach:**
Using `<script>` in the index.html file as described in the docs.

Got this error:
`ConfigError: Couldn't resolve parser "solidity-parse". Plugins must be explicitly added to the standalone bundle.`

Code Sample using `<script>` (for import is almost the same, just replace the `prettierPlugins `for `solidityPlugin`:
```
const formattedCode = await prettier.format(sourceCode, {
	parser: "solidity-parse",
	plugins: [prettierPlugins],
	singleQuote: false,
	printWidth: 200,
	tabWidth: 4,
	useTabs: false,
	bracketSpacing: false,
});
```

Is there anything I am missing? Thanks in advance.

Janther:
For the browser solution, if the file has been properly loaded, `prettierPlugins` should be populated with the solidity plugin already.
You could log the values of `solidityPlugin` and `prettierPlugins`, this might help find the issue.

For the NPM approach, I see you are using `prettier/standalone`.
You could try to load also our standalone file:

```Javascript
import prettier from 'prettier/standalone';
import solidityPlugin from 'prettier-plugin-solidity/dist/standalone.cjs';
```

Another approach is to tell `vite` to always load the browser friendly files.
I haven't used `vite` but having a quick look at their documentation, they support `resolve.alias` in their configuration.

Could you see if this works:
```Javascript
resolve: {
  alias: [
    { find: 'prettier', replacement: 'prettier/standalone' },
[...]

PR #943, opened 2023-11-04
Build now uses `prettier/standalone`

Janther:
With this change users won't have to uses aliases when building applications for the browser.

This feature was by 2 users (#942 and another one in the Telegram thread), having issues when building an application for the browser.

Before this fix, webpack was generating code that used `require('prettier')`. This would lead to requiring the use of aliases in the webpack configuration so `'prettier'` would be resolved to `'prettier/standalone'`.

With this change, users will still be able to build for node using `commons`, `esm`, and building for the browser without the need for aliases.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -36,6 +36,8 @@ jobs:
         run: npm install
       - name: Build
         run: npm run build
+      - name: Build test app
+        run: npm run build:test
       - name: Run tests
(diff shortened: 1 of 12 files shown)


fvictorio (approved):
I haven't tested this manually, but looks good. Maybe we should publish a `dev` version first and ask those users that asked about this to give it a try?

PR #949, opened 2023-11-08
ready to publish

Janther:
In this PR:

- Adding tests to support solidity 0.8.22
- bumping @solidity-parser/parser to 0.16.2
- bumping solc to 0.8.23
- bumping package version to 1.2.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ For more details and please have a look at [Prettier's documentation](https://pr
 
 ### Creating a package for the Browser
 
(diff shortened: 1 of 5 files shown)


Janther:
Had to remove the windows machine because it won't install a dependency of `solc`. I made them aware in this issue https://github.com/ethereum/solc-js/issues/715

Will keep track of this.

Janther:
I reverted the bump on solc@0.8.23 so we can safely publish a version that runs on all machines

PR #953, opened 2023-11-19
Experimental ternaries

Janther:
I really liked Prettier's [`experimentalTernaries`](https://prettier.io/blog/2023/11/13/curious-ternaries) and would like to propose it for this plugin as well.

I took the same examples from prettier's tests.

There is one detail that makes this approach slightly different and that is that the length of the indentation in javascript is 2 and that makes the true and false expressions align.

```Javascript
condition ?
  trueExpression
: falseExpression;
```

with 4 spaces in solidity the code looks like this

```Solidity
condition ?
    trueExpression
: falseExpression;
```

I could force this indentation to be 2 only for this one case but would love to see how the community reacts to this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -206,6 +206,19 @@ You might have a multi-version project, where different files are compiled with 
 | ------- | --------------------- | ---------------------- |
 | None    | `--compiler <string>` | `compiler: "<string>"` |
 
+### Experimental Ternaries
+
+Mimicking prettier's [new ternary formatting](https://prettier.io/blog/2023/11/13/curious-ternaries) for the community to try.
+
+Valid options:
+
+- `true` - Use curious ternaries, with the question mark after the condition.
+- `false` - Retain the default behavior of ternaries; keep question marks on the same line as the consequent.
+
+| Default | CLI Override               | API Override                    |
+| ------- | -------------------------- | ------------------------------- |
+| false   | `--experimental-ternaries` | `experimentalTernaries: <bool>` |
+
 ## Integrations
 
 ### Vim
--- a/src/binary-operator-printers/logical.js
+++ b/src/binary-operator-printers/logical.js
@@ -5,13 +5,19 @@ const { group, line, indent } = doc.builders;
 const groupIfNecessaryBuilder = (path) => (document) =>
   path.getParentNode().type === 'BinaryOperation' ? document : group(document);
 
-const indentIfNecessaryBuilder = (path) => (document) => {
+const indentIfNecessaryBuilder = (path, options) => (document) => {
   let node = path.getNode();
   for (let i = 0; ; i += 1) {
     const parentNode = path.getParentNode(i);
     if (parentNode.type === 'ReturnStatement') return document;
     if (parentNode.type === 'IfStatement') return document;
     if (parentNode.type === 'WhileStatement') return document;
+    if (
+      options.experimentalTernaries &&
+      parentNode.type === 'Conditional' &&
+      parentNode.condition === node
+    )
+      return document;
     if (parentNode.type !== 'BinaryOperation') return indent(document);
     if (node === parentNode.right) return document;
     node = parentNode;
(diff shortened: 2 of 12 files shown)


fvictorio (approved):
This looks great! I won't have time to review the implementation, but I reviewed the snapshots and played with it a bit locally.

pcaversaccio:
As a quick feedback. I like that you incorporate ternaries into this plugin, but I disagree with the following logic:

image

To increase the readability of ternaries, in particular, nested ternaries (see e.g. my PR [here](https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3854)), I usually add parentheses. This might be an opinionated rule but it enforces explicitness IMO. Thus, I don't agree with the current implementation that it removes those. I will disable this rule for the moment being.

Janther:
Thank you for your feedback.

That behaviour is the outcome of a feature that created a bug that created a feature that in your particular opinion is actually a bug.

Initially we wanted to wrap the condition in parentheses if it was too long.

```Solidity
// Input

very.long.condition() ? foo : bar

// Not desired output
very
  .long
  .condition() ?
  foo
: bar

// Desired output
(
  very
    .long
    .condition()
) ?
  foo
: bar
```

but this created a possible second formatting since that parenthesised condition, would also break
```Solidity
// second formatting
(
  (
    very
      .long
      .condition()
  )
) ?
  foo
: bar
```
Thus I decided to remove parentheses from the condition only since we know these always will be a single boolean and only add them back if the condition happens to break.

This is also the behaviour Prettier has in Javascript for ternaries.

As a nice side effect this also dealt with double parentheses `((condition)) ? true : false` that could be added by mistake in a lousy copy paste scenario.

After new year, I'll have another look at this and give it a look and see if there is a compromise that can be implemented.

Also I had a look at your PR and we only remove parentheses from the condition not the consequent which is where you added the parentheses, so those should be ok.

Finally those parentheses are not needed for that scenario anymore since this plugin would format them as follows.

```Solidity
// Input
bytes32 b = proofFlags[i] ? leafPos < leavesLen ? leaves[leafPos++] : hashes[hashPos++] : proof[proofPos++];

// experimentalTernaries: true
bytes32 b =
    proofFlags[i] ?
        leafPos < leavesLen ?
            leaves[leafPos++]
        :   hashes[hashPos++]
    :   proof[proofPos++];

// experimentalTernaries: false
bytes32 b = proofFlags[i]
    ? leafPos < leavesLen
        ? leaves[leafPos++]
        : hashes[hashPos++]
    : proof[proofPos++];
```

pcaversaccio:
Thanks for the elaboration. It creates a particular issue for me in the context of my `CreateX` code: https://github.com/pcaversaccio/createx/blob/4db1d64a9b3331821cdca791bac5037ac2fddba5/src/CreateX.sol#L910:

```solidity
// For the non-pseudo-random cases, the salt value `salt` is hashed to prevent the safeguard mechanisms
// from being bypassed. Otherwise, the salt value `salt` is not modified.
guardedSalt = (salt != _generateSalt()) ? keccak256(abi.encode(salt)) : salt;
```

Having the parentheses here makes it much more readable since you understand it's a condition. Since I don't want to change any metadata due to `prettier-plugin-solidity`, I had to remove the prettier for the affected files (see https://github.com/pcaversaccio/createx/pull/59)). Unfortunately, I can't configure this behaviour (the same holds true for `JS` `prettier`, that's why I had to put it into the `.prettierignore` file).

Janther:
Actually, this behaviour should have only affected `experimentalTernaries`, I just published a [quick fix](https://www.npmjs.com/package/prettier-plugin-solidity/v/1.3.1) that leaves parentheses untouched if `experimentalTernaries` is `false`.

pcaversaccio:
> Actually, this behaviour should have only affected `experimentalTernaries`, I just published a [quick fix](https://www.npmjs.com/package/prettier-plugin-solidity/v/1.3.1) that leaves parentheses untouched if `experimentalTernaries` is `false`.

Awesome, thanks a lot! Can confirm it fixes it. Merry Christmas!

PR #961, opened 2023-12-08
Improving printing performance

Janther:
Removing `group`s that will always break:

- UncheckedStatement
- EnumDefinition
- ModifierDefinition

Removing `group` that just affect a closing bracket `)` or `]` that will be outside the `tabWidth`

- TupleExpression

Member Access Chain is already properly `group`ed

- MemberAccess

Just reordering the `group` `indent` calls:

- FunctionDefinition
- FunctionTypeName

Prefer using `shouldBreak` at the `group` level instead of choosing between `hardline` and `line`
- IfStatement

The rest is wrapping strings together whenever possible.
Finally I rebased some functions that would `push` and `concat` into a variable and returned an built array instead.

### DIFF
--- a/src/common/printer-helpers.js
+++ b/src/common/printer-helpers.js
@@ -8,7 +8,7 @@ import {
 const { group, indent, join, line, softline, hardline } = doc.builders;
 
 export const printComments = (node, path, options, filter = () => true) => {
-  if (!node.comments) return '';
+  if (!node.comments) return [];
   const document = join(
     line,
     path
--- a/src/common/printer-helpers.js
+++ b/src/common/printer-helpers.js
@@ -37,35 +37,27 @@ export const printComments = (node, path, options, filter = () => true) => {
 };
 
 export function printPreservingEmptyLines(path, key, options, print) {
-  const parts = [];
-  path.each((childPath, index) => {
+  return path.map((childPath, index) => {
     const node = childPath.getValue();
-    const nodeType = node.type;
 
-    if (
-      // Avoid adding a hardline at the beginning of the document.
-      parts.length !== 0 &&
+    return [
+      // Only attempt to prepend an empty line if `node` is not the first item
+      index > 0 &&
       // LabelDefinition adds a dedented line so we don't have to prepend a
       // hardline.
-      nodeType !== 'LabelDefinition'
-    ) {
-      parts.push(hardline);
-    }
-
-    parts.push(print(childPath));
-
-    // Only attempt to append an empty line if `node` is not the last item
-    if (
+      node.type !== 'LabelDefinition'
+        ? hardline
+        : '',
+      print(childPath),
+      // Only attempt to append an empty line if `node` is not the last item
       !isLast(childPath, key, index) &&
+      // Append an empty line if the original text already had an one after the
+      // current `node`
       isNextLineEmpty(options.originalText, options.locEnd(node) + 1)
-    ) {
-      // Append an empty line if the original text already had an one after
-      // the current `node`
-      parts.push(hardline);
-    }
+        ? hardline
+        : ''
+    ];
   }, key);
-
-  return parts;
 }
 
 // This function will add an indentation to the `item` and separate it from the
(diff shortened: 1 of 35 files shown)


src/common/printer-helpers.js
```diff
@@ -8,7 +8,7 @@ import {
 const { group, indent, join, line, softline, hardline } = doc.builders;
 
 export const printComments = (node, path, options, filter = () => true) => {
-  if (!node.comments) return '';
+  if (!node.comments) return [];
```
Janther, line 11: if this functions always returns an array, then we can always ask for `comments.length` instead of `comments?.length`

fvictorio:
This results in some code changes in the OpenZeppelin repository:

```diff
diff --git a/contracts/access/extensions/AccessControlDefaultAdminRules.sol b/contracts/access/extensions/AccessControlDefaultAdminRules.sol
index ef71a648..73a8b66b 100644
--- a/contracts/access/extensions/AccessControlDefaultAdminRules.sol
+++ b/contracts/access/extensions/AccessControlDefaultAdminRules.sol
@@ -183,7 +183,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      */
     function defaultAdminDelay() public view virtual returns (uint48) {
         uint48 schedule = _pendingDelaySchedule;
-        return (_isScheduleSet(schedule) && _hasSchedulePassed(schedule)) ? _pendingDelay : _currentDelay;
+        return _isScheduleSet(schedule) && _hasSchedulePassed(schedule) ? _pendingDelay : _currentDelay;
     }
 
     /**
@@ -191,7 +191,7 @@ abstract contract AccessControlDefaultAdminRules is IAccessControlDefaultAdminRu
      */
     function pendingDefaultAdminDelay() public view virtual returns (uint48 newDelay, uint48 schedule) {
         schedule = _pendingDelaySchedule;
-        return (_isScheduleSet(schedule) && !_hasSchedulePassed(schedule)) ? (_pendingDelay, schedule) : (0, 0);
+        return _isScheduleSet(schedule) && !_hasSchedulePassed(schedule) ? (_pendingDelay, schedule) : (0, 0);
     }
 
     /**
diff --git a/contracts/governance/extensions/GovernorTimelockCompound.sol b/contracts/governance/extensions/GovernorTimelockCompound.sol
index 77cf369f..2837ca80 100644
--- a/contracts/governance/extensions/GovernorTimelockCompound.sol
+++ b/contracts/governance/extensions/GovernorTimelockCompound.sol
@@ -40,8 +40,8 @@ abstract contract GovernorTimelockCompound is Governor {
         ProposalState currentState = super.state(proposalId);
 
         return
-            (currentState == ProposalState.Queued &&
-                block.timestamp >= proposalEta(proposalId) + _timelock.GRACE_PERIOD())
+            currentState == ProposalState.Queued &&
+                block.timestamp >= proposalEta(proposalId) + _timelock.GRACE_PERIOD()
                 ? ProposalState.Expired
                 : currentState;
     }
diff --git a/lib/forge-std b/lib/forge-std
index c2236853..eb980e1d 160000
--- a/lib/forge-std
+++ b/lib/forge-std
@@ -1 +1 @@
-Subproject commit c2236853aadb8e2d9909bbecdc490099519b70a4
+Subproject commit eb980e1d4f0e8173ec27da77297ae411840c8ccb
```

Maybe we are missing tests for ternaries whose condition is inside parentheses?

Also, I used hyperfine to benchmark this change, and these are the results (again, in OZ) before and after the change:

```
// Before
  Time (mean ± σ):     11.554 s ±  0.315 s    [User: 16.764 s, System: 3.959 s]
  Range (min … max):   10.990 s … 11.857 s    10 runs


// After
  Time (mean ± σ):     11.458 s ±  0.246 s    [User: 16.539 s, System: 3.937 s]
  Range (min … max):   11.024 s … 11.824 s    10 runs
```

So I don't think it has an impact, at least in that repo. That doesn't mean I'm against this change though, because it simplifies things anyway.

Janther:
Thank you for this,
after this [comment](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/953#issuecomment-1868340445) addressing the same issue I decided to keep the original behaviour although in the future if prettier decides to adopt `experimentalTernaries` as the default formatting choice and we decide to follow that choice we will need to sometimes add parentheses the thing will be when to remove them.

But that's for the future.

Janther:
All of the changes of this PR will be included in the  TypeScript #968 step by step migration so I decided to close this one in favor to that one

PR #968, opened 2023-12-30
Migration to typescript Step 1

Janther:
Renaming all files from `js` to `ts` makes the review process quite difficult.
In order to make the migration to typescript as easy and as clean as possible, I split it in a few steps.

important changes in this step:

- install typescript packages and add necessary configuration to build, load, and test.
- add final types
- migrated all files that `git` would recognise as a file renaming and not a delete `x.js` and create `x.ts`
- in the remaining `js` files changed the minimal amount of logic for the whole package to still work.

All tests still passing


Janther:
Closing because `ts-node` has issues with `node 16` and mixing `esm` and `cjs`, which our tests need.

Janther:
Not doing since the `antlr` parser is deprecated

PR #989, opened 2024-02-16
Add ZORA as a user of prettier solidity

iainnash:
ZORA is a happy user of the prettier solidity package and we'd be happy to be in the readme.


fvictorio:
Thanks!

PR #1033, opened 2024-08-16
Necessary tweaks to the current formatting

Janther:
So there won't be many critical changes to the tests when moving to Slang.


src/nodes/AssemblyStackAssignment.js
```diff
@@ -1,7 +1,12 @@
+import { doc } from 'prettier';
+
+const { hardline } = doc.builders;
+
 export const AssemblyStackAssignment = {
   print: ({ node, path, print }) => [
     path.call(print, 'expression'),
-    ' =: ',
+    hardline,
```
Janther, line 8: In Slang (and also in the Solidity documentation)
The stack assignment starts at the `=:` and it's a statement by itself.
to pair it with the previous statement, would cause a very particular exception in the slang implementation which I believe is not very productive, specially for a feature that was removed on version `0.5.0`.

src/nodes/FunctionTypeName.js
```diff
@@ -24,16 +24,19 @@ const stateMutability = (node) =>
...
-    ')',
-    indent(
-      group([
-        visibility(node),
-        stateMutability(node),
-        returnTypes(node, path, print)
-      ])
-    )
-  ]
+  print: ({ node, path, print }) =>
```
Janther, line 27: FunctionTypeName did not follow the same rules we struggled for so long regarding how a function declaration should group and indent.
Here we are picking up this case we forgot.

src/nodes/HexLiteral.js
```diff
@@ -1,11 +1,11 @@
...
 import { printString } from '../common/util.js';
 
-const { join, line } = doc.builders;
+const { join, hardline } = doc.builders;
 
 export const HexLiteral = {
   print: ({ node, options }) => {
     const list = node.parts.map((part) => `hex${printString(part, options)}`);
-    return join(line, list);
+    return join(hardline, list);
```
Janther, line 9: the only reason to have a single string declared as a list of strings is when you want to visually display them in separate lines. It makes no sense otherwise.

src/nodes/ModifierDefinition.js
```diff
@@ -38,16 +32,22 @@ const override = (node, path, print) => {
...
+  return [path.call(print, 'body')];
 };
 
 export const ModifierDefinition = {
   print: ({ node, path, print }) => [
     'modifier ',
     node.name,
     modifierParameters(node, path, print),
-    group(indent([virtual(node), override(node, path, print)])),
+    group(
```
Janther, line 44: ModifierDefinition also did not follow the same rules we discussed for FunctionDefinitions.

src/prettier-comments/language-js/comments.js
```diff
@@ -257,21 +257,51 @@ function handleIfStatementComments(
     precedingNode === enclosingNode.trueBody &&
     followingNode === enclosingNode.falseBody
   ) {
-    if (precedingNode.type === "ExpressionStatement") {
-      addTrailingComment(precedingNode, comment);
+    if (followingNode.type === "Block") {
```
Janther, line 260: In slang we have way more control over comments and inspired by Prettier's decision of no having comments between the `trueBody` and `falseBody` we are gonna put it at the beginning of the `falseBody`. This also removes a lot to edge case logic in `ExpressionStatement.js` and `IfStatement.js`

fvictorio (approved):

PR #1047, opened 2024-09-27
Infer version

Janther:
The version inference works as follows:

1. Slang is error-tolerant, so we can parse using any version.
2. We read the pragma statements from the parsed output.
3. We use the pragma statements to figure out the latest supported version.

Simple enough but when developing I encountered cases where:

- slang would still throw when parsing
- there are no pragma statements
- the pragma statements is not a supported version

Therefore, to mitigate errors we calculate milestone versions which are the first and last patch versions of any minor version supported (`0.8.26`, `0.8.0`, `0.7.6`, `0.7.0`, `0.6.12`, `0.6.0`, `0.5.17`, `0.5.0`, `0.4.26`, `0.4.11`).

Starting from the latest version, we attempt the original steps.
If there are errors or no pragma statement, we attempt with the next version (this will add a processing tax on older syntaxes)
If the pragma statement is not a supported version (e.g. `0.8.27`), we assume the latest supported version.

Another albeit ridiculous thing that could happen is that a developer could add a random comment.
```Solidity
pragma solidity ^0.5.0 /* ^0.6.0 */ ^0.7.0;
```

To address this we use our own AST structure, to fetch the comments and replace them manually with empty space from the unparsed version range.


Janther:
I think that once we recover the pragmas we should go for the `minSatisfyingVersion` instead of the `maxSatisfyingVersion`. Any thoughts?

fvictorio (approved):

PR #1061, opened 2024-11-03
storing offsets in a Map instead of passing them down through an array

Janther:
the initial approach that moved offsets between parent and child has been replaced by an approach that stores and consumes the offset based in the CST id.

This id has been available since slang 0.18.2.

the main file to review here is `slang-utils/metadata.ts` since everything else is just the removal of the passing along of the offsets.


fvictorio:
I only skimmed the code but I don't think I can offer a valuable review here. Feel free to merge without my approval. But if you have specific things you are not sure about, let me know and I'll check those.

PR #1063, opened 2024-11-03
removing strip-comments and using prettier to clean pragma versions

Janther:
Aware this might be an overkill but the possible issue is really difficult to solve with other means (see tests).

I have a different approach to get rid of the `strip-comments` dependencies if this gets rejected.


src/slang-utils/create-parser.ts
```diff
@@ -25,15 +32,48 @@ const milestoneVersions = Array.from(
...
+          locEnd
+        } as PrettierParser
+      },
+      printers: { ['slang-ast']: { print: slangPrint } }
+    }
+  ],
+  parser: 'slangPragma'
+};
+
+const query = Query.parse(
```
mattiaerre, line 64: I noticed you moved `query` out of the `tryToCollectPragmas` function, but you only use it there; should we move it back to its original location?

tests/unit/slang-utils/create-parser.test.js
```diff
@@ -83,21 +83,31 @@ describe('inferLanguage', function () {
     {
       description:
         'should use the latest version if the range is outside the supported versions',
-      source: `pragma solidity ^0.8.27;`,
+      source: `pragma solidity ^0.9.27;`,
       version: latestSupportedVersion
+    },
+    {
+      description: 'broken by new lines, whitespace and comments',
```
mattiaerre, line 90:
```suggestion
      description: 'broken by new lines, whitespaces, and comments',
```

mattiaerre (approved):

src/slang-utils/create-parser.ts
Janther: this way the query is computed only once and not every time it tryToCollectPragmas is called

mattiaerre (approved):
🚀

fvictorio:
I think I'd rather not support very weird pragma versions, like the one in the new test, and rather check that we throw a decent error when that happens (throw "couldn't parse version pragma in file <file>" and crash).

Janther:
I made a new PR I had prepared if this one was too ambitious and complicated maintenance

PR #1066, opened 2024-11-14
Remove strip-comments dependency

Janther:
This one is a trimmed down PR #1063.
It's cleaner and more straight forward but also not being as aggressive.

The error we throw if presented with the weird pragmas is "Couldn't infer any version from the ranges in the pragmas.".


fvictorio (approved):
I pushed a commit adding the filename to the error message. If you are ok with that, feel free to merge.

Janther:
just added the case where there is no filepath. Also passing the options object as the changes I'm working on will use that and doesn't affect the current implementation

PR #1107, opened 2025-03-23
Antlr/Slang tests

Janther:
comparing slang and antler formats when testing so we are aware when there is a breaking change.


tests/config/run-format-test.js
```diff
@@ -65,6 +64,22 @@ const testsWithAstChanges = new Map(
   }),
 );
 
+const unstableAntlrTests = new Map(
```
fvictorio, line 67: Very minor, but "unstable" in the context of prettier means something else, which is how we use it below: it means that formatting it twice doesn't produce the same result as formatting it once. This name seems to say that some tests are unstable when using the ANTLR parser, but I think these are instead tests that result in a different output when using ANTLR vs Slang.

Maybe we can use something like `divergentTests` or `antlrMismatchTests`, or something like that.

PR #1167, opened 2025-06-11
Using `node` instead of `getNode()`

Janther:
it has a cleaner type for the type checks, and the standard in `prettier`'s parsers.
using the same approach, `getNode(2)` was replaced by `grandparent`.
I also added some really quick refactors that I found when replacing these cases and don't really merit another PR.

### DIFF
--- a/src/slang-printers/print-binary-operation.ts
+++ b/src/slang-printers/print-binary-operation.ts
@@ -44,7 +44,7 @@ export const binaryIndentRulesBuilder =
   (shouldIndent: (node: BinaryOperation) => boolean) =>
   (path: AstPath<BinaryOperation>) =>
   (document: Doc): Doc => {
-    for (let i = 2, { node } = path; ; i += 2) {
+    for (let i = 2, node = path.node; ; i += 2) {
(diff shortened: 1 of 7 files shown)


fvictorio (approved):
I don't agree with some of the refactors here (e.g., I think destructuring + aliases like `const { foo: bar } = qux` are generally less readable than `const bar = qux.foo`), but since this is mostly a personal preference, I'm ok with merging this PR as is.

Janther:
happy to review these destructures in most cases I use the alias in a function signature where I don't have access to the `=` and renaming everything to the property name would look weird:

```TypeScript
function printComment({ node: comment }) {
  if (isComment(comment)) {
    return comment.print();
  }

  throw new Error(`Not a comment: ${JSON.stringify(comment)}`);
}
```

And the other place where I use the alias is in a `for...of` loops with a similar reason.

```Typescript
    for (const { variant: attribute } of this.attributes.items) {
      if (
        typeof attribute !== 'string' &&
        attribute.kind === NonterminalKind.ModifierInvocation
      ) {
        attribute.cleanModifierInvocationArguments();
      }
    }
```

PR #1175, opened 2025-06-20
better handling of empty array from `inferLanguage`

Janther:
There are 2 cases in which we receive an empty array from `inferLanguage`.

- the range is outside of the supported versions.
- there are ranges in the pragmas that don't intersect.

with this in mind we are defaulting to the latest supported version and return a more detailed error.

also in this PR:

I could not parse any nightly builds even when using specific versions of nightly builds of `solc`. Therefore I'm removing this test.

https://github.com/NomicFoundation/slang/issues/1346

### DIFF
--- a/src/slang-utils/create-parser.ts
+++ b/src/slang-utils/create-parser.ts
@@ -21,6 +21,15 @@ function parserAndOutput(
   };
 }
 
+function createError(
+  { parseOutput }: { parseOutput: ParseOutput },
+  reason: string
+): Error {
+  return new Error(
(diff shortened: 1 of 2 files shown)


Janther:
the issue I'm trying to fix here is when there is a new solidity version and `slang` takes some time in releasing the support and then we take some time on adapting to this, and finally the user installs our latest release.
With the current solution we won't even attempt to format even if the newest solidity version has no syntax changes or the developer is not using the newest syntax.

PR #1178, opened 2025-06-23
only break import deconstruction if the list of items is longer than 1

Janther:
by looking at openzeppelin's  long imports I noticed that forcing to break a single deconstructed import into multiple lines, only makes the file longer and the most important part is the imported object.

This makes files with multiple imports where some of them break and some don't a bit confusing to read. That's why I think we could review the import rule so only deconstructions longer than one item will break.

**This will force the `openzeppelin-contracts` package to re-write some of the contracts that were just updated when they upgraded to `prettier-plugin-solidity@2.0.0` because of this reason, which is undesirable but I believe is an important tweak.**


fvictorio (approved):
I was on the fence about this change, but apparently Prettier core does the same, so yeah, let's do it.

PR #1217, opened 2025-07-24
using Slang API to ignore trivia nodes

Janther:
mainly future proofing in case Slang adds a new trivia `kind` in future versions.


Janther:
closing in favour of #1235

PR #1230, opened 2025-08-10
Singleton comments

Janther:
instead of gathering comments at each node and collecting them at the `updateMetadata` step, we store them in a single place and retrieve them at the end of the parsing step.

### DIFF
--- a/src/slang-nodes/PositionalArgumentsDeclaration.ts
+++ b/src/slang-nodes/PositionalArgumentsDeclaration.ts
@@ -27,9 +27,11 @@ export class PositionalArgumentsDeclaration extends SlangNode {
 
     // We need to check the comments at this point because they will be removed
     // from this node into the root node.
+    // Since we are collecting comments in a different array, we have to query
+    // the ast directly for possible block comments.
     this.isEmpty =
       this.arguments.items.length === 0 && // no arguments
-      !this.comments.some((comment) => isBlockComment(comment)); // no block comments
+      !ast.cst.children().some(({ node }) => isBlockComment(node)); // no block comments
   }
 
   print(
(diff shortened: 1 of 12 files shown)


src/slang-nodes/PositionalArgumentsDeclaration.ts
```diff
@@ -27,9 +27,11 @@ export class PositionalArgumentsDeclaration extends SlangNode {
 
     // We need to check the comments at this point because they will be removed
     // from this node into the root node.
+    // Since we are collecting comments in a different array, we have to query
+    // the ast directly for possible block comments.
     this.isEmpty =
       this.arguments.items.length === 0 && // no arguments
-      !this.comments.some((comment) => isBlockComment(comment)); // no block comments
+      !ast.cst.children().some(({ node }) => isBlockComment(node)); // no block comments
```
Janther, line 34: this is the only node that has a change that is not removing `updateMetadata`

src/slang-nodes/SourceUnit.ts
```diff
@@ -21,10 +21,6 @@ export class SourceUnit extends SlangNode {
     this.members = new SourceUnitMembers(ast.members, options);
 
     this.updateMetadata(this.members);
-
-    // Because of comments being extracted like a russian doll, the order needs
-    // to be fixed at the end.
-    this.comments = this.comments.sort((a, b) => a.loc.start - b.loc.start);
```
Janther, line 27: comments are handled at the `slangSolidityParser` level

src/slang-printers/print-comments.ts
```diff
@@ -13,9 +17,10 @@ function isPrintable(comment: Comment): boolean {
 }
 
 export function printComments(
-  path: AstPath<AstNode>,
+  path: AstPath<StrictAstNode>,
   options: ParserOptions<AstNode>
 ): Doc[] {
+  if (typeof path.node.comments === 'undefined') return [];
```
fvictorio, line 23: This is a weird pattern to see nowadays. `typeof x === 'undefined'` made sense in old javascript where `x` could be undeclared, but I think today you just use `x === undefined`, even more so in typescript.

fvictorio (approved):
Pre-approving with a minor comment.

PR #1305, opened 2025-11-08
Add knip

fvictorio:
This PR adds [knip](https://knip.dev/), a tool to find unused code. I've used in a couple of projects and it works quite well.

The most important thing to review here are the files I'm adding to the `ignore*` fields in the `knip.json` config. I'm erring on the side of not deleting things, since I don't know whether they are used or not.

I also added `npm run knip` to the `lint` script. Knip is quite fast, so this shouldn't make that script much slower. But if you think this is annoying, we can change that and only run it in the CI.

I'm creating this PR on top of https://github.com/prettier-solidity/prettier-plugin-solidity/pull/1296, since that's the last big refactor.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -10,7 +10,5 @@
     "interface": true,
     "type": true
   },
-  "tags": [
-    "-lintignore"
-  ]
+  "tags": ["-lintignore"]
 }


knip.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://unpkg.com/knip@5/schema.json",
+  "ignore": ["src/nodes/*.js"],
+  "ignoreFiles": [
+    "tests/format/RespectDefaultOptions/respect-default-options.js",
+    "tests/integration/node.test.js"
+  ],
```
fvictorio: The difference between `ignore` and `ignoreFiles` is not super obvious:

- `ignore` means "don't report unused code inside this file".
- `ignoreFiles` means "don't report the whole file as unused"

PR #1500, opened 2026-04-06
using the latest `print` signature

Janther:
Since Prettier 2.3.0 the `print` argument for the `print` function has a new [signature](https://prettier.io/docs/plugins#print) and since it was backwards compatible, we never noticed 😅, however we can now take advantage of this and have a way less polluted environment and smaller build.

```TypeScript
function print(
  // Path to the AST node to print
  path: AstPath,
  options: object,
  // Recursively print a child node
  print: (selector?: string | number | Array<string | number> | AstPath) => Doc,
): Doc;
```

This opens the possibility to shorten the call `path.call(print, 'selector')` to `print('selector')` making our code much cleaner.

This PR has 2 steps:

1. Change all instances of `path.call(print, 'selector')` to `print('selector')`
2. Remove all unused references to `path` that are no longer needed

For the second step, was necessary to change the signature of the `print` function of each node so the `path` parameter would come after the `print` parameter and therefore removable when not used.

```TypeScript
  // Before
  print(
    path: AstPath<AdditiveExpression>,
    print: PrintFunction,
    options: ParserOptions<AstNode>
  ): Doc {
    return printAdditiveExpression(this, path, print, options);
  }

  print(path: AstPath<ArrayExpression>, print: PrintFunction): Doc {
    return ['[', path.call(print, 'items'), ']'];
  }

  // After
  print(
    print: PrintFunction,
    path: AstPath<AdditiveExpression>,
    options: ParserOptions<AstNode>
  ): Doc {
    return printAdditiveExpression(this, path, print, options);
  }
  
  print(print: PrintFunction): Doc {
    return ['[', print('items'), ']'];
  }
```

I tested these changes with prettier 3.0.0 and it was supported back then.

Caveats:
The prettier team hasn't updated their TypeScript declarations for this function so at one point I had to cast it to the old signature that they provide. [I made a PR to their project.](https://github.com/prettier/prettier/pull/19014)

There is a loss of type checking as `path.call(print, 'name')` checks that the current node has a node called `name` and that the `print` function is equiped to receive said node to be printed.
However after having seen @fisker comment, prettier checks just before printing that the node exist on runtime (**only when we are using prettier's on development**).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -12,7 +12,10 @@ import { locEnd, locStart } from './slang-utils/loc.js';
 import { hasPrettierIgnore } from './slang-utils/has-prettier-ignore.js';
 
 import type {
+  AstPath,
+  Doc,
   Parser,
+  ParserOptions,
   Printer,
   RequiredOptions,
   SupportLanguage
@@ -77,7 +80,12 @@ const slangPrinter: Printer<PrintableNode | undefined> = {
   handleComments,
   isBlockComment,
   massageAstNode,
-  print: slangPrint,
+  print: slangPrint as (
+    path: AstPath<PrintableNode | undefined>,
+    options: ParserOptions<PrintableNode | undefined>,
+    print: (path: AstPath<PrintableNode | undefined>) => Doc,
+    args?: unknown
+  ) => Doc,
   hasPrettierIgnore,
   printComment
 };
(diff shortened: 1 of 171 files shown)


fisker:
Core plugins forbid pass non-node to print https://github.com/prettier/prettier/blob/3675b1cfd55f94a47f4b0aef60249b1d0a686d80/src/main/create-print-pre-check-function.js#L3

fvictorio (approved):

PR #1501, opened 2026-04-07
delegating printing nodes with `prettier-ignore` to prettier

Janther:
`hasPrettierIgnore` has been documented since prettier 3.7.0 but has been in the `types` and in the code since at least prettier 3.0.0

this way we won't need to maintain `ignoreComments`.

I tested this with prettier 3.0.0 and it was supported back then.


fvictorio (approved):
Nice!

PR #1506, opened 2026-04-09
switching `eslint-plugin-import` with `eslint-plugin-import-x`

Janther:
for compatibility with eslint 10. It's been 2 months since the release of eslint 10 and `eslint-plugin-import` has been very slow in the adoption of the new linter.

`eslint-plugin-import-x` started as a fork that now, not only works with eslint 10, but also has stopped supporting older versions of node in favour of faster processing.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -1,7 +1,7 @@
 import globals from 'globals';
 import typescriptEslint from '@typescript-eslint/eslint-plugin';
 import tsParser from '@typescript-eslint/parser';
-import eslintImport from 'eslint-plugin-import';
+import eslintImportX from 'eslint-plugin-import-x';
 import path from 'node:path';
 import { fileURLToPath } from 'node:url';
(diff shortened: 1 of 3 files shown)


fvictorio:
Hmm, I'm not sure about this one. Two months is not that long of a delay for an OSS project to make a major upgrade (they are probably unpaid maintainers) and I'd rather not add new dependencies if we can avoid it given all the recent supply chain attacks.

What's the trade-off here? What do we lose if we don't have the very latest version of eslint and this plugin?

Janther:
It can wait.

it's supposed to be faster, built on rust, and not supporting really outdated node and eslint versions.

all very good things but it affects us only in one step of our development process which is adding a new import to a file.

if it keeps blocking us for too long, I'd suggest again this change though.

fvictorio:
Should we close this one for the time being, for bookkeeping?

Janther:
sure

PR #1539, opened 2026-05-04
EndOfChain decided on construction

Janther:
to avoid iterating through parents multiple times when printing.


fvictorio (approved):
I have to say I prefer the previous code because it wasn't stateful. But tests pass, so I'm ok with merging it if you think it's worth it.

Janther:
Strictly, it should be a `readonly` attribute set up in the constructor. I just don't know yet how it will affect the rest of the codebase by adding this very particular exception 😅
I'll keep cooking this one as I believe it's building towards something.

Janther:
Closing since it doesn't really improve anything, other than perhaps the file size.

At first glance I thought there was a possible of having the loop go multiple times over the same nodes and therefore the other approach would remove that, but upon a deeper analysis, this doesn't happen.

PR #1660, opened 2026-08-05
destructuring `pretier.doc.builders` and `prettier.util` only once

Janther:
slight performance improvement at the beginning of the runtime and smaller file size since every time we destructure an object, we have to name the attribute and we can't minimize that.


fvictorio (commented):
Very interesting, and makes sense.

fvictorio (approved):

PR #1673, opened 2026-08-13
Miscelaneous

Janther:
Got to play with the eslint plugin unicorn and even though I don't think we should adopt it now, I found these suggestions really nice for our code maintenance.

### DIFF
--- a/src/slang-nodes/PositionalArgumentsDeclaration.ts
+++ b/src/slang-nodes/PositionalArgumentsDeclaration.ts
@@ -30,7 +30,7 @@ export class PositionalArgumentsDeclaration extends SlangNode {
     // the ast directly for possible block comments.
     this.isEmpty =
       this.arguments.items.length === 0 && // no arguments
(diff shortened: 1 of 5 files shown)


src/slang-nodes/PositionalArgumentsDeclaration.ts
```diff
@@ -30,7 +30,7 @@ export class PositionalArgumentsDeclaration extends SlangNode {
     // the ast directly for possible block comments.
     this.isEmpty =
       this.arguments.items.length === 0 && // no arguments
-      !ast.cst.children().some(({ node }) => isBlockComment(node)); // no block comments
+      ast.cst.children().every(({ node }) => !isBlockComment(node)); // no block comments
```
Janther, line 33: I like how this reads in natural English:

```diff
- Not some child is BlockComment
+ Every child is not BlockComment
```

fvictorio (approved):

PR #1716, opened 2026-09-18
dependabot now also tracks `github/actions`

Janther:
Bonus, dependabot now also updates the dependencies in `package.json`


fvictorio (approved):
Nice!

